PR #28, opened 2015-04-23
Added support for domain-based logins

luiswebmercado:
I really loved this API and I'm using it in a project here in my company that requires this type of configurations...

I does not impact in "non-domain-based" logins.

I added some code into Wikimate's login method and change the README file, adding info about how to inform and where to inform the domain name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,14 +32,17 @@ to login to that wiki.
 
 ```php
 include 'globals.php';
-	
+
 $api_url = 'http://example.com/api.php';
 $username = 'bot';
 $password = 'password';
-	
+
 try
 {
 	$wiki = new Wikimate($api_url);
+
+    // You can pass the domain name either
+    // $wiki->login($username,$password, $domainName)
 	if ($wiki->login($username,$password))
 		echo 'Success: user logged in.' ;
 	else {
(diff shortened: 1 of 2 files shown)


Wikimate.php
```diff
@@ -62,6 +62,13 @@ public function login( $username, $password ) {
 			'lgpassword' => $password,
 			'format' => 'json'
 		);
+
+		// If $domain is informed, sets the correspondant detail in the request information array
+		if( $domain )
```
samwilson: Should this be a more robust check, e.g. `$domain !== NULL` or `is_string($domain)`?

waldyrious:
@samwilson, the original author seems to be MIA. I could merge and make the change you propose. Which of your suggestions do you think would be preferable?

README.md
```diff
@@ -32,14 +32,17 @@ to login to that wiki.
...
 $username = 'bot';
 $password = 'password';
-	
+
 try
 {
 	$wiki = new Wikimate($api_url);
+
+    // You can pass the domain name either
+    // $wiki->login($username,$password, $domainName)
```
waldyrious: Indentation needs adjusting.

Wikimate.php
```diff
@@ -62,6 +62,13 @@ public function login( $username, $password ) {
...
+		{
+			$details['lgdomain'] = $domain;
+
```
waldyrious: Excess whitespace here.

Wikimate.php
samwilson: I reckon `is_string($domain)` would be most readable.

luiswebmercado:
Hey! Sorry the miss attention in this topic, I did not see the notification about these messages.
I agree with everything discussed here.

Thanks for the merge.

PR #30, opened 2016-09-29
Add missing property; correct get->debugRequestsConfig call; some...

Xymph:
…comment and formatting fixes

Started using this project for a bot on DoomWiki.org and found some improvements and fixes to make in this and the next few commits.

I'm new to GitHub so I'm grouping my edits into a few separate commits rather than one bulky update. Hopefully this makes accepting/merging them easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Then, download Wikimate, and initialise it by running `composer install` (or
 To use Wikimate within another project, you can add it as a composer dependency
 by adding the following to your `composer.json` file:
 
-    "hamstar/Wikimate": "0.10.0"
+    "hamstar/Wikimate": "0.10.1"
 
 ## Usage
 
@@ -224,6 +224,12 @@ $array_result = $wiki->edit( $data );
 Both methods return an array of the MediaWiki API result.
 
 ## Changelog
+
+### Version 0.10.1
+
+* Supports optional domain at authentication
+* Bug fix: pass return value in setSection
+* Bug fix: correct call to debugRequestsConfig
 
 ### Version 0.10.0
 
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -28,7 +28,7 @@ class Wikimate {
 	
 	/**
 	 * Create a new Wikimate object.
-	 * 
+	 *
 	 * @return Wikimate
 	 */
 	function __construct( $api ) {
@@ -46,7 +46,7 @@ class Wikimate {
 	 */
 	protected function initRequests() {
 		$this->session = new Requests_Session( $this->api );
-		$this->useragent  = "Wikimate ".self::VERSION." (https://github.com/hamstar/Wikimate)";
+		$this->useragent = "Wikimate ".self::VERSION." (https://github.com/hamstar/Wikimate)";
 	}
 
 	/**
@@ -124,7 +124,6 @@ class Wikimate {
 		
 		//Logger::log( "Logged in" );
 		return true;
-		
 	}
 	
 	/**
@@ -243,6 +242,7 @@ class Wikimate {
 		
 		return unserialize( $apiResult );
 	}
+	
 	public function getError() {
 		return $this->error;
 	}
@@ -277,9 +277,9 @@ class WikiPage {
 	
 	/**
 	 * Constructs a WikiPage object from the title given and adds
-	 * a wikibot object
+	 * a Wikimate object
 	 * @param string $title name of the wiki article
-	 * @param WikiBot $wikibot WikiBot object
+	 * @param Wikimate $wikimate Wikimate object
 	 */
 	function __construct( $title, $wikimate ) {
 		$this->wikimate = $wikimate;
@@ -290,7 +290,6 @@ class WikiPage {
 			echo "Invalid page title - cannot create WikiPage";
 			return null;
 		}
-		
 	}
 	
 	/**
@@ -481,17 +480,16 @@ class WikiPage {
 		}
 		
 		return $this->text; // Return the text in any case
-		
 	}
 	
 	/**
 	 * Returns the section requested, section can be the following:
 	 * - section name (string:"History")
 	 * - section index (int:3)
-	 * 
+	 *
 	 * @param mixed $section the section to get
 	 * @param boolan $includeHeading false to get section text only
-	 * @return string wikitext of the section on the page 
+	 * @return string wikitext of the section on the page
 	 */
 	function getSection( $section, $includeHeading = false ) {
 		// Check if we have a section name or index
@@ -509,13 +507,12 @@ class WikiPage {
 			$text = substr( $this->text, $offset );
 		}
 		
-		// Whack of the heading if need be
+		// Whack off the heading if need be
 		if ( !$includeHeading && $offset > 0 ) {
 			$text = substr( $text, strpos( trim( $text ), "\n" ) ); // Chop off the first line
 		}
 		
 		return $text;
-		
 	}
 	
 	/**
@@ -548,7 +545,6 @@ class WikiPage {
 		}
 		
 		return $sections;
-		
 	}
 	
 	/*
@@ -571,7 +567,7 @@ class WikiPage {
 			'md5' => md5( $text ),
 			'bot' => "true",
 			'token' => $this->edittoken,
-			'starttimestamp' => $this->starttimestamp 
+			'starttimestamp' => $this->starttimestamp
 		);
 		
 		// Set options from arguments
@@ -604,7 +600,7 @@ class WikiPage {
 			$data = array(
 				'prop' => 'info',
 				'intoken' => 'edit',
-				'titles' => $this->title 
+				'titles' => $this->title
 			);
 			
 			$r = $this->wikimate->query( $data );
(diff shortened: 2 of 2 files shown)


waldyrious:
@Xymph thanks for the patch! Before you make further commits, I'd ask you to split the current one into one with just the cleanup changes, and another with the fix, if you don't mind. Are you able to check this branch locally, edit the commits and force-push? Let me know if you need any help with that.

Xymph:
@waldyrious Thanks for the prompt follow-up. I've so far been using the web GUI rather than a local git clone, and in the GUI I don't see how to make the changes you're asking for. So yes, please provide further guidance how to split the commit. Btw, there are two code fixes (one very minor), which one are you referring to as "the fix"?

waldyrious:
> I've so far been using the web GUI rather than a local git clone, and in the GUI I don't see how to make the changes you're asking for. So yes, please provide further guidance how to split the commit. 

I assumed so :) indeed the web UI is quite convenient and I prefer it whenever possible, but some tasks are still only possible in the command line (though they've been improving in this regard recently). So to split the commits, here's what I'd do:

- Clone your fork of the repo: `git clone https://github.com/Xymph/Wikimate.git && cd Wikimate`
- Checkout the branch of this PR: `git checkout patch-1`
- Uncommit the last commit (changes remain in the filesystem, ready to commit again): `git reset HEAD^`
- Interactively choose which parts (hunks) of the changes to add for commit. Use the "s" to split hunks further, "y" to add them to the commit, "n" to keep them unstaged (to be added in a later commit): `git add -p`
- Commit the staged subset of the changes: `git commit -m "some comment and formatting fixes"`
- Commit the rest of the changes: `git commit -am "Add missing property; correct get->debugRequestsConfig call"`
- Update your branch in github (`--force` is required since you're not just adding new commits, but also removing an existing one): `git push --force`

That's it! the PR here in the Wikimate repo should be automatically updated after you do that.

I know it seems like a lot of steps, but each one should be relatively simple, I hope. And sorry if some of my wording sounds condescending -- I don't know how comfortable you are with these git operations, so I thought I'd rather risk providing too many details than too little 😄. In any case, let me know if you need anything clarified :)

> Btw, there are two code fixes (one very minor), which one are you referring to as "the fix"?

Yes, sorry for the ambiguity.

Xymph:
@waldyrious No problem, I did some CLI work with git awhile ago, but it's a little rusty and I wasn't familiar with the reset/hunk-add process, so your detailed instructions are certainly welcome.
Most steps went fine, including the commits. Had some brief trouble with the push until I found the "remote set-url origin" fix with "Xymph@" in the URL.

waldyrious:
Awesome, glad I could help. Are you planning on sending further changes as separate PRs, or were you thinking about updating this PR with more commits? In case you don't know, adding new commits to an existing PR branch is possible to do straight from the web interface (with the limitation that each commit can only edit one page at a time).

Xymph:
Well, that is something I've been wondering about: I have another bug fix, two missing @param lines, and a little update of the destructor; but also a more extensive update of the getText section processing code. The former looks like an update to 0.10.1 to me (together with the first two commits); the latter to 0.11.0 -- so both meaning README.md updates too.

What's the best way to proceed and at which point(s) would you want to do merge(s)?

waldyrious:
It's up to you, the commits will look the same in the repo regardless of what branch they come from. I suppose it's easier to send separate PRs, but let me know.

As for version updates, I don't feel comfortable making that call, hopefully others will chime in. I am just a PR submitter like you who @hamstar gave contributor access to the repo after a while 😄.

Xymph:
Ok, I made a few separate cli commits to hopefully make acceptance of each/all easier.

The version numbers are a bit weird anyway, as the comments above class WikiPage also have one, but it's still at 0.5 - yet there's no VERSION constant in that class. The one in class Wikimate refers to semver.org which recommends to increment the last number in case of bug fixes. So I guess my fixes qualify for an increment, although they were made in both classes so I'm not sure whether WikiPage also needs an increment.

waldyrious:
We should definitely harmonize the version information throughout the project. It's nice that @hamstar specified SemVer -- it is a reliable, predictable system to follow, and I agree a patch increase makes sense for these changes. Can you also update the version in the WikiPage class' documentation?

Also, pinging @samwilson and @MhdSyrwan in case they want to chime in.

Xymph:
Sure, done. In that case I also added mention of last July's update to the readme. One might argue that each class could keep its own version (e.g. in case the classes grow so much they get split into separate files), but then I'm not sure how to handle this cleanly in the Changelog. Input from others would be welcome indeed.

waldyrious:
I don't think it makes sense to maintain separate versions for each class. I was suggesting changing WikiPage to 0.10.1 as well, and keeping things consistent from then on.

Thanks for the update on Wikipedia! Indeed, thanks for these thoughtful changes and the discussion :)

Xymph:
My pleasure, I've used stuff off GitHub before, so it's good to now give back and contribute.

If there should be no distinct version numbers per class -- which would be my preference too -- then I think it would be better to have just one in the documentation sections of the .php file (i.e. above Wikimate), and delete the one above class WikiPage. Otherwise there's a chance that a future edit results in a discrepancy again. Agreed?

waldyrious:
Agreed. I didn't suggest it before because I thought it could be a common convention in PHP circles or something :)

Xymph:
I've done plenty work in PHP and more than one version number per project, let alone per file (in any language), is... rather unusual. ;) But perhaps @hamstar had a good reason for it at the time.

So what's the next step?

waldyrious:
I would actually create a non-doxygen-style comment block at the top of the file to put the author, version and since metadata, and leave the class-specific comments to information related to their inner workings and APIs.

Also, I am very tempted to suggest an interactive rebase to reorder / combine / reword some of these commits into a cleaner and more logical history, but I don't want to abuse your patience, so let me know if you're up for that. It's worth knowing how to do it, though ;)

waldyrious:
@Xymph I guess my growing enthusiasm for git shows at times, which is why I'm being cautious to not give the impression of a requirement on my part, but rather of an availability to assist in case *you* are interested. Any commit cleanup that would be done here isn't really important as you say, especially since the commits are already quite well separated. Mind you, though, that I don't plan to squash them together, but create a merge commit instead, which will preserve in the repo's history the gradual, logically-independent steps that the changes in this PR entail.

My suggestion, should you accept it, would be to (1) shorten the commit messages, so that github and other interfaces with limited width can handle them properly; (2) combine commits that make sense as a logical unit, e.g. d817ae9 and 34c899a; (3) optionally reordering the commits into whatever you, as the author, feel is the most logical sequence.

That said, I repeat: I have no issues whatsoever in merging this branch with the commits as-is, and I want to make sure you don't feel that I'm asking you to do this. If you're ok with this PR as it is, we can certainly merge already.

Xymph:
No worries (I'm not Australian but I like the phrase :) ), I accept (1) and (2).  Please talk (err, write) me through the steps. Apart from forgetting the $echo change in 7885a3b, the commit order is something I did think about beforehand, and seems good enough for my first PR.

waldyrious:
Cool! Ok, so here's what you'd do: 

Assuming you still have the local clone of your fork in your machine, `cd` into it and `git checkout patch-1` to go to this branch, in case you weren't on it already. Normally a rebase would, as the name says, change the base of a branch (i.e. transplant it elsewhere), but your branch is [already sitting at the top of the master branch](https://github.com/hamstar/Wikimate/network). Instead, what this'll do is "rebase" it to the same place, but using the interactive option, which allows you to change the commits during the rebase. So, to start the interactive rebase, do `git rebase -i master`.

You will then be presented with a list of commits in your branch, in chronological order (oldest at the top). This will use your system's default in-terminal editor, usually vim. If so, change to insert mode (press `i`) and move the cursor next to the commits you will want to change the wording of.

Erase the "pick" that appears next to those commits, and write "reword" instead. Do this for every commit that will be reworded (but don't change the actual commit message there). Then, exit the insert mode with `esc`, and move the cursor to a commit you want to combine into another one. Press `d` to cut that line, move the cursor to the line of the commit it is to be merged into, ans press `p` to paste it below it. Then, change again into insert mode and replace "pick" by "fixup" in the commit to be fused into the other.

(By the way, there's a handy guide for all of this at the bottom of the commit list, so you won't have to refer back to this text during the rebase.)

With this done, you can save and exit the editor (`esc`, `:wq`, `enter` in vim), which will start the rebase. It will proceed automatically until it reaches the first commit marked for rewording. The instructions should be printed in the command line at that time: `git commit --amend -m "new commit message"`, then `git rebase --continue`. You will be prompted like this for each commit marked for rewording, and at the end the rebase will be concluded.

Finally, to update the PR, just do a force-push.

ps - once more, I apologize in case I'm providing more instructions than you need (vim stuff, etc). I'm following the "better safe than sorry" maxim :)

Xymph:
Been using vi[m] since... well, let's not give my age away that easily :) but that's okay. And `d` should be `dd`. ;)
I shortened and fixed up what I thought made sense, although the rebase process continued to be interactive and put me right back into vim for the message edits (so no need for amend and continue commands).
If you agree it's good like this, then we could wait a bit longer for those paged people to chime -- or wrap it up, that's up to you.

waldyrious:
Hahah, good catch with the `dd`, sorry about that! 😅 And regarding the interactive rebase process, I must have mixed up its behavior with "reword" vs. "edit" -- again, sorry for that.

The commits look absolutely pristine now, thanks so much for putting up with my perfectionism! I think since nobody has commented even after being mentioned, we can just go ahead and merge the PR. 🎉

PR #32, opened 2016-10-02
fix documentation for consistency and readability

waldyrious:
This should be fairly uncontroversial, but putting it out as a PR in case anyone has comments.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -1,10 +1,16 @@
 <?php
+/// =============================================================================
+/// Wikimate is a wrapper for the MediaWiki API that aims to be very easy to use.
+/// 
+/// @version    0.10.1
+/// @copyright  SPDX-License-Identifier: MIT
+/// =============================================================================
+
 /**
  * Provides an interface over wiki API objects such as pages.
  *
- * @author Robert McLeod
- * @since December 2010
- * @version 0.10.1
+ * @author  Robert McLeod
+ * @since   December 2010
  */
 class Wikimate {
 	
@@ -12,7 +18,7 @@ class Wikimate {
 	const SECTIONLIST_BY_INDEX = 2;
 
 	/**
-	 * @var string The current version number (conforms to http://semver.org/).
+	 * @var  string  The current version number (conforms to http://semver.org/).
 	 */
 	const VERSION = '0.10.1';
 
@@ -20,7 +26,7 @@ class Wikimate {
 	protected $username;
 	protected $password;
 
-	/** @var Requests_Session */
+	/** @var  Requests_Session */
 	protected $session;
 	protected $useragent;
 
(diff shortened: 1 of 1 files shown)


Wikimate.php
```diff
@@ -196,8 +205,9 @@ public function query( $array ) {
 
 	/**
 	 * Performs a parse query to the wiki API.
-	 * @param array $array array of details to be passed in the query.
-	 * @return array unserialized php output from the wiki API.
+	 *
+	 * @param   array  $array  Array of details to be passed in the query.
```
Xymph: All other param lines have no dot at the end - trim them in these two lines too?

Wikimate.php
```diff
@@ -356,47 +372,52 @@ function destroy() {
...
+	 * Getter functions
 	 *
 	 */
 	
 	/**
-	 * Gets the text of the page.  If refesh is true then this method will
-	 * query the wiki api again for the page details
-	 * @param boolean $refresh true to query the wiki api again
-	 * @return string the text of the page
+	 * Gets the text of the page. If refesh is true,
```
Xymph: Please fix the old typo in refesh while you're at it.

Wikimate.php
```diff
@@ -356,47 +372,52 @@ function destroy() {
...
+	 * then this method will query the wiki api again for the page details
```
Xymph: Append a period to this sentence.

Xymph:
@waldyrious Besides the specific changes proposed above, the last WikiPage method `delete` doesn't have a Doxygen block at all. And since the code tests `$reason`'s existence, I'm thinking it should default to null in the parameter list as well: `delete( $reason = null )`

What were you saying about perfectionism again? :wink:

Wikimate.php
```diff
@@ -130,8 +137,8 @@ public function login( $username, $password, $domain = NULL ) {
 	/**
 	 * Sets the debug mode
 	 *
-	 * @param boolean $debugMode true to turn debugging on
-	 * @return Wikimate this object
+	 * @param  boolean $debugMode true to turn debugging on
```
waldyrious: Good catch 👍

Wikimate.php
waldyrious: Makes sense.

Wikimate.php
waldyrious: Ah, I did catch a typo elsewhere, but this one eluded me. Will fix :)

Wikimate.php
waldyrious: 👍

waldyrious:
Great eye for detail @Xymph :) I believe I've applied all the fixes you suggested (and a couple more that I missed previously). Please take another look!

waldyrious:
Thanks for the review 👍

Issue #31, opened 2016-09-30
Improvements to section processing


Xymph:
The section processing in getText (from the preg_match_all down) has some problems:

1. The regex (rightly marked with a TODO) can be improved so that it matches headers with a symmetric number of ='s, and those that happen to have spaces behind them.
1. The last section (both by index and by name) isn't given a length in the sections arrays. getSection works around this but it's not elegant.
1. Pages with duplicate section headers (see e.g. [Item](https://doomwiki.org/wiki/Item)) result in fewer byName entries than byIndex, and incorrect offset / negative length values in both.

I have code changes for these ready, but would like to understand (from a commit organizational perspective) whether these are better applied in a separate branch & PR or could also hitchhike along in the ongoing patch-1 of PR #30. Thoughts?

waldyrious:
I personally would prefer these as a separate PR. The thread in #30 is long enough already :)

Xymph:
That was my thinking too, so thanks a lot for wrapping that one up.

I'm assuming I can continue to use my patch-1 branch for the commits in this issue? Or should it be deleted now, and new one started?
If the latter, I can use the Delete button in the previous PR, but how do I then sync my local git tree?
As an aside, I've been wondering why the Branch dropdown list on the main page never listed patch-1?

Further, can I simply create a new branch from the command line and commit changes referencing this issue number? Or should I create a new pull request? I must admit still being a little confused about GitHub's mix of web GUI and command-line approaches to the same repo's, and I have found little in their Help and Guides sections to clarify the latter.

waldyrious:
> I'm assuming I can continue to use my patch-1 branch for the commits in this issue? Or should it be deleted now, and new one started?

Normally people tend to delete the branch and start a new one, but that's just a convention. In semantic terms, it makes more sense for new branches to be based off master (i.e. the merge commit that has `patch-1` and the previous commit on `master` as parents, rather than the last commit of `patch-1` before the merge). But in technical terms it doesn't really make much of a difference (well, it may make eventual rebases easier, but I never attempted that.)

[...]

PR #33, opened 2016-10-02
Improvements to section processing

Xymph:
See #31 item 1. Rationale for this change:
Allowing optional trailing spaces is needed, or a section header can be inadvertently skipped. Optional leading spaces don't matter as the pattern isn't anchored there.
Using a back-reference to the first set of ='s insures a symmetric number of them is matched, just like the Wiki engine does.
A minimizer `?` to the inner `.*` makes sense in this situation.

After making these changes I thought to look in the API source, and found in [this file](https://phabricator.wikimedia.org/diffusion/MW/browse/master/includes/EditPage.php) at line 2086 a nearly identical regex:
`$hasmatch = preg_match( "/^ *([=]{1,6})(.*?)(\\1) *\\n/i", ...);`
So I think we're good now and the TODO can be removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,10 @@ Both methods return an array of the MediaWiki API result.
 
 ## Changelog
 
+### Version 0.11.0
+
+* Improved section processing in getText
+
 ### Version 0.10.1
 
 * Supports optional domain at authentication
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -471,9 +471,9 @@ class WikiPage {
 				$currIndex = 0;
 				$currName  = 'intro';
 				
-				foreach ( $sections as $i => $section ) {
+				foreach ( $sections as $section ) {
 					// Get the current offset
-					$currOffset = strpos( $this->text, $section );
+					$currOffset = strpos( $this->text, $section, $this->sections->byIndex[$currIndex]['offset'] );
 					
 					// Are we still on the first section?
 					if ( $currIndex == 0 ) {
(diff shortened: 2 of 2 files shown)


Xymph:
See item 2. While `getSection` checks for the existence of lengths (`if ( isset( $length ) )... else...`), a user of `getSectionOffsets` may rightly expect the last section to have a length value too, without having to code a test for it like the class does internally. This change defines them by Index and by Name.

I'm considering to remove the test from `getSection` as a consequence. Opinions?

waldyrious:
All the changes look good to me, and I agree the now redundant test in `getSection` can be removed.

Additionally, could you move the comments at the end of lines to above them? This should make some of the longer lines a bit more manageable.

Xymph:
See item 3. When multiple sections have the same name, the second (and subsequent) sector(s) need(s) to be distinguished in the byName array. For this I've used the same approach as the Wiki engine does in anchoring them, by appending `_2`, `_3` and so on. See also the parse output for the [example page](https://doomwiki.org/w/api.php?action=parse&page=Item&contentmodel=wikitext&prop=sections).

Further, the strpos'es that search for each next section need to use the current one as starting point, not the start of the page, as that resulted in offset errors and negative lengths. With the committed changes, the values in both arrays match, and the offsets from `getSectionOffsets` for the example page correlate with the above parse output (apart from a 1-byte discrepancy for each non-ASCII /UTF-8 character, such as `×`, which is counted as 1 character by parse and 2 bytes in getText).

Xymph:
@waldyrious Thanks for the initial feedback, will do. There's one more round above.
As mentioned before I'd like to update the version to 0.11.0 (and update the README), for which I need to pull your just committed doc changes to master into my branch. I'm gonna be lazy :) and ask, could you provide the git command(s) to perform that?

waldyrious:
I appreciate the detailed write-up. Your original issue and the third commit's code & message were enough to figure out what was the intent, but explicit in-context description is always a plus.

As for the branch update: you will need to do a rebase (an actual one this time :D), like this:

- Change to the master branch:  
  `git checkout master`
- Sync your local repo's master branch with the upstream changes:  
  `git pull upstream master`  
  (assuming you have called this repo's remote "upstream")
- Move back into your branch:  
  `git checkout section-improvements`
- Rebase it into the new tip of `master` (this should happen automatically without conflicts):  
  `git rebase master`
- Update your fork on github, and thus this PR as well:  
  `git push --force`  
  (assuming you've used `push --set-upstream` before)

That should be it, but let me know if any of this doesn't work as expected :)

Xymph:
Thanks again, that worked fine. I fixup-ed the length test into the earlier commit (and, accidentally, one comment fix), then added the readme/version/comment updates in the last commit. If that's all, then this branch is ready for merge back.

waldyrious:
Yep, looks good to me! Nice work 👍

PR #34, opened 2016-10-03
split README into separate files

waldyrious:
This should make the content easier to locate and read.


waldyrious:
LICENSE.md wasn't added, actually. There already existed a LICENCE file, I just renamed it directly (without a PR) when I did a couple minor fixes to it a right before sending this PR: 3579e53

Xymph:
Ah right, just realized. Thanks anyway, that 'C' was bugging me a bit too. :wink:

waldyrious:
And yes, I'm planning on adjusting the releases (create tags and possibly releases for the old versions mentioned in CHANGELOG.md, if that won't screw up the ordering), and at that point I'll create a new release and update the README as well 👍

Xymph:
Just noticed an old typo: in USAGE, `getAllSections(false, WikiPage:SECTIONLIST_BY_NAME)` should have two colons in the class constant, `WikiPage::SECTIONLIST_BY_NAME`.

Speaking of class constants, the ones in class Wikimate aren't used. Since that functionality exists (and belongs) in WikiPage, I think they should be removed from Wikimate.

Lastly, I found three more lowercase mentions of 'api' in the doxygen blocks, above Wikimate query and WikiPage getText (twice). Might wanna fix those before creating the release.

Xymph:
I found a smal bug in the section processing, so I'll create a new issue for that in which the above Wikimate.php edits can hitchhike along. Then you don't need to include code edits in this documentation ticket.

waldyrious:
Thanks for the detailed review. I would rather not make any changes to the text in this PR, because they get obscured in the larger edit. But I'll certainly do those fixes in a subsequent PR. In fact, lets merge this so that those changes can be made afterwards :)

PR #37, opened 2016-10-03
Define lengths for page without sections (fixes #35)

### LINKED ISSUE #35: Page without sections leaves length undefined
Because ef19eec removed the `isset( $length )` test, a page without sections results in a PHP Notice on that `substr` statement, as the `length` fields in the section arrays are left undefined for the first/intro section.

Also, I found three more lowercase mentions of 'api' in the doxygen blocks, above Wikimate query and WikiPage getText (twice).
And the Wikimate constants SECTIONLIST_BY_* can be removed, those belong in class WikiPage only.

Xymph:
See #35. In case of no sections, this sets the lengths of the 0/intro "section" to the page length.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -466,7 +466,11 @@ class WikiPage {
 			$this->sections->byIndex[0]['offset']      = 0;
 			$this->sections->byName['intro']['offset'] = 0;
 			
-			if ( !empty( $m[0] ) ) {
+			if ( empty( $m[0] ) ) {
+				// Define lengths for page without sections
+				$this->sections->byIndex[0]['length']      = strlen( $this->text );
+				$this->sections->byName['intro']['length'] = strlen( $this->text );
+			} else {
 				// Array of section names
 				$sections = $m[0];
 				
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -186,7 +186,7 @@ class Wikimate {
 	}
 	
 	/**
-	 * Performs a query to the wiki api with the given details
+	 * Performs a query to the wiki API with the given details
 	 *
 	 * @param   array  $array  Array of details to be passed in the query
 	 * @return  array          Unserialized php output from the wiki API
@@ -410,9 +410,9 @@ class WikiPage {
 	
 	/**
 	 * Gets the text of the page. If refresh is true,
-	 * then this method will query the wiki api again for the page details.
+	 * then this method will query the wiki API again for the page details.
 	 *
-	 * @param   boolean  $refresh  True to query the wiki api again
+	 * @param   boolean  $refresh  True to query the wiki API again
 	 * @return  string             The text of the page
 	 */
 	function getText( $refresh = false ) {
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -160,10 +160,10 @@ class Wikimate {
 	 * Get or print the Requests configuration.
 	 *
 	 * @param   boolean  $echo  Whether to echo the options
-	 * @return  array           Options if $echo is FALSE
+	 * @return  array           Options if $echo is false
 	 * @return  boolean         True if options have been echoed to STDOUT
 	 */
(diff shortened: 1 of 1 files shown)


waldyrious:
Note: If you add "(fixes #35)", "(closes #35)", (or a few other variants thereof) to the title of the PR, the corresponding issue will automatically be closed once the PR is merged :)

Xymph:
Ok, wasn't sure when and where to do that: on the commit(s) or the PR.
You're welcome to review and, if approved, merge this one.

waldyrious:
Looks good, I would just prefer the comment on 41a8e96 to be a tad more descriptive (i.e. explaining what that `if` condition means, and that we're creating a "virtual" section encompassing the whole page).

Xymph:
Hmm, the "if empty" follows logically from the preg_match_all and the comments above the offset and length setting statements. The area before the first section is called [lead or intro](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Lead_section) and if no section header is present (no preg match results) then the intro is, indeed, the entire page. But it is and already was a part of the sections list. The offsets for intro were already defined, that commit merely defined their lengths.
So in context, I don't think further commentary is needed. Does that make sense?

waldyrious:
I don't think I see your point. All code can be followed logically and interpreted directly (with more or less effort) to figure out what it does, but comments in natural language are precisely to avoid the need to "run" the code in your head.

For instance, `$m` is a pretty opaque variable name; why force the code reader to go looking for its previous occurrences, only to figure out in what condition the if block runs?

Likewise, the comment "define lengths for page without sections" is unnecessarily vague: define the length to what? Since there's a comment there already, it's a waste not to use that space to provide the full information of what the code it describes does.

Of course, this isn't a blocker, and if you insist that any further documentation is excessive, I won't object. I'm just trying to make sure we're understanding each other clearly; if so, we can just agree to disagree.

Xymph:
Alright, `$m` is indeed overly curt -- but it was already there. My point was (also) that my new comments segued onto the existing ones, which also weren't overly descriptive. To improve them, I need to tweak/rewrite quite a few more lines, and I was under the impression that commits for bug fixes should be minimal in number of affected lines, in order to facilitate quick reviews and not expand the scope of the commit beyond the bug fix. That's something I learned at my last job (where we switched from svn to git a few months before I left).

So with that restriction loosened :) I made a bunch more edits and think the comment flow is clear now. Also, I found an all-caps literal that [should](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#25-keywords-and-truefalsenull) to be lowercased. Lastly, I noticed I wasn't used to the spacing style around parentheses yet in some #33 commits, and fixed those.

That style is [non-standard](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#5-control-structures) but is what hamstar started out with, so I'm not sure we should touch that (in a separate PR).

waldyrious:
> commits for bug fixes should be minimal in number of affected lines, in order to facilitate quick reviews and not expand the scope of the commit beyond the bug fix.

That is an argument I can certainly agree on. I wouldn't object to local improvements that weren't comprehensive, per the [boy-scout rule](http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule), but keeping PRs' scope from expanding is an equally good principle. What I took issue with was the principle I perceived to be behind "follows logically from the preg_match_all and the comments above" and "in context, I don't think further commentary is needed", which I disagree with for the reasons I tried to explain above. Given the scope expansion concern, which I agree is valid, feel free to remind me in future PRs should I make similar requests :)

> I made a bunch more edits and think the comment flow is clear now. 

Agreed; I'll merge now. Thanks also for the additional formatting fixes. 👍

Xymph:
I didn't know it was called that, but I do like the boy-scout rule, and usually apply it when the need/opportunity arises, but within reasonable limits. Glad we agree after all. :)

PR #41, opened 2016-10-05
Test for unknown section (fixes #39)

### LINKED ISSUE #39: getSection: unknown $section gives PHP Notices
The `$coords = ` statements don't first test whether `$section` exists in the `byIndex` or `byName` arrays. An unknown `$section` gives a PHP Notice on the assignment, and two more on the `substr()` a bit later.

Xymph:
Hmm, I accidentally branched 39-fix-notices off my other branch instead of master. @waldyrious can you delete it and the PR? Sorry for the inconvenience.

In the 39-fix-unknown-section branch the tests return `false` for unknown sections, a common practice in many built-in PHP functions, and the user can test for that using `===`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,18 @@
 ## Changelog
 
+### Version 0.11.1
+
+* Bug fix: handle unknown section parameter correctly in `WikiPage::getSection()`
+
 ### Version 0.11.0
 
-* Improved section processing in getText
+* Improved section processing in `WikiPage::getText()`
 
 ### Version 0.10.1
 
 * Supports optional domain at authentication
-* Bug fix: pass return value in setSection
-* Bug fix: correct call to debugRequestsConfig
+* Bug fix: pass return value in `WikiPage::setSection()`
+* Bug fix: correct call to `Wikimate::debugRequestsConfig()`
 
 ### Version 0.10.0
 
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -531,13 +531,22 @@ class WikiPage {
 	 *
 	 * @param   mixed    $section         The section to get
 	 * @param   boolean  $includeHeading  False to get section text only
-	 * @return  string                    Wikitext of the section on the page
+	 * @return  string                    Wikitext of the section on the page,
+	 *                                    or false if section is undefined
 	 */
 	function getSection( $section, $includeHeading = false ) {
 		// Check if we have a section name or index
 		if ( is_int( $section ) ) {
+			if ( !isset( $this->sections->byIndex[$section] ) )
+			{
+				return false;
+			}
 			$coords = $this->sections->byIndex[$section];
 		} else if ( is_string( $section ) ) {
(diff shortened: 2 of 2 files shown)


waldyrious:
The fork point of a branch shouldn't be an issue, since you can always rebase it.

Also, the branch can be deleted, but not by me, for the reasons I mentioned in https://github.com/hamstar/Wikimate/issues/31#issuecomment-250960023: branches created in forks and used as basis for PRs aren't pulled into the upstream repo, but instead only internal git references are created to allow operations like diffs, comments, and to preserve history in case the fork is deleted. So if you want to delete the branch, you must do it yourself, in https://github.com/Xymph/Wikimate/branches.

As for the PR, I'm not sure which one you're talkign about. It seems to me you never finished the process to create the PR based on the `39-fix-notices` branch... or am I missing something?

waldyrious:
As for the changes themselves, they seem fine since they make the behavior more explicit, but I wonder if someone could be relying on the old behavior. If so, we should add a note in the changelog for this change.

Xymph:
> The fork point of a branch shouldn't be an issue, since you can always rebase it.

Didn't know, but then there are far more things about git that I don't know yet than that I do. :wink: I guess I really should read up and refresh/expand my knowledge, but I haven't had time yet (read: writing bot scripts is more fun :sunglasses: ). I noticed my master branch was still behind, but at least found [this recipe](https://help.github.com/articles/syncing-a-fork/) to take care of that. And any closed branches are gone now, thanks for that pointer.

> As for the PR, I'm not sure which one you're talking about.

Yes, I was prompted to create the PR and didn't, so later the prompt disappeared.

> I wonder if someone could be relying on the old behavior. If so, we should add a note in the changelog for this change.

For "old behavior" we'd need to go back to before all the recent commits, i.e. f0558f0 . I tested it and there a undefined section causes the entire page to be returned. That hardly seems something that a user would rely on.
But it is a bug fix nonetheless, so a changelog entry and a version increment would be in order.

waldyrious:
> I guess I really should read up and refresh/expand my knowledge, but I haven't had time yet

If you ever feel in the mood to learn a bit about how git works, I can't recommend [Git For Ages 4 and Up](https://www.youtube.com/watch?v=1ffBJ4sVUb4) enough. I've read plenty of introductions, tutorials, guides, explainers, tl;drs, and various portions of the official documentation (much of which is actually very good reference material), but this one beats all of those by a mile in terms of intuitiveness of the explanation and creating a lasting mental model which will help any further learning and explanation feel much more natural and logical.

Thanks for the doc updates. This looks ready to 🚢! Merging now.

Xymph:
Thanks for the link. It took a number of sessions to sit through the whole thing, and the camera work and editing left, uhh, room for improvement :wink:, but it was very enlightening indeed.

After #52 I don't expect any further code changes soon (famous last words :smile:), so you could consider creating that 0.11.0 release some time soonish.

waldyrious:
Glad to hear the video was useful :) As for the release, I will try tackling this this weekend, or sooner if I get the chance.

PR #43, opened 2016-10-06
Prevent PHP Notices in delete/setText (fixes #42)

### LINKED ISSUE #42: delete & setText methods give PHP Notices
`Wikimate::delete` runs unserialize on an object, generating a PHP Notice.
`WikiPage::delete` tests for the `'delete'` element in the return array that may not exist (in case of an error, like permission denied), resulting in another PHP Notice.

Similarly, `setText` tests for `$r['edit']['result']` which also doesn't exist in case of an error.

Xymph:
The `isset`s insure no PHP Notices can occur in those if statements.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -68,8 +68,7 @@ class Wikimate {
 		);
 
 		// If $domain is provided, set the corresponding detail in the request information array
-		if( is_string( $domain ) )
-		{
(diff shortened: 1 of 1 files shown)


waldyrious:
Nice fix, and almost flawless separation of commits ;)

LGTM 👍

Issue #38, opened 2016-10-04
newSection: effect of $name parameter is not consistent


Xymph:
Discovered this while writing a script that uses this method (am I really the first? :)).
$name is passed onto $summary in the setSection call, and ends up as summary in the page's editing history. So it should be named $summary accordingly.

I considered the possibility that the intent really was to pass the section header, but that makes no sense. Then setSection or setText would have a $name parameter too, and somewhere the header and body would need to be combined, but for that the section depth (number of ='s) would also be needed. So no: $text needs to include everything the user wants to put into the section, including header (or not).

**Edit**: I may be mistaken, looking into it further.

Xymph:
The pertaining MediaWiki API [edit parameters](https://www.mediawiki.org/wiki/API:Edit#Parameters) behave in a way I didn't expect:

If `section=new` and `sectiontitle` is undefined (as it is in Wikimate) then `summary` is used as new section header with a fixed level of 2 (wrapped in '=='). Additionally, `summary` is embedded in a standardized edit summary, `/* summary */ new section`, which is then rendered in the page history as usual: _→summary: new section_ (with hyperlinked arrow to the section anchor, and greyed section name).

In contrast, when editing an existing section, `summary` is used verbatim. But if a new section is created and the `sectiontitle` parameter **is** included (edited my local WikiPage to test that), then `summary` is used verbatim in the history too!
I find this rather counter-intuitive, but I guess the API designers had a good reason for it.

Anyway, `newSection`'s `$name` parameter is therefore correctly named as it currently stands.

We could debate whether to add a `sectiontitle` parameter to `newSection`, but that would then also need to be passed through `setSection` and `setText`. This would grant the user more control over the summaries in a page history (in the bot scripts I'm writing I'd like to flag them as "Automated edits"), but change the parameter list of multiple methods. Maybe that's not worth it for new sections, which seems to me is a task far less common for bots than text and section edits. What do you think, @waldyrious ?

waldyrious:
I see. This behavior reflects what the actual mediawiki UI does: when creating a new section (which normally only happens in discussion pages), you don't get a summary field, only a "subject" field above the edit box, which becomes the title of the section, and part of the automatic edit summary, as you describe.

> then `summary` is used verbatim in the history too! I find this rather counter-intuitive

Did you mean to say that `sectiontitle` is used in the history as the edit summary? Otherwise I'm not sure I understand why using `summary` as the edit summary (i.e. in the history) is counter-intuitive.

[...]

PR #45, opened 2016-10-06
Add new findSection method and use it in setText (fixes #40)

### LINKED ISSUE #40: setSection/setText: allow specifying section by name
Having written a few bot scripts using Wikimate by now, this experience leads to yet another issue (btw, how can issues be labeled as Bug, Feature request, etc?).

One can obtain a section's content via its name using `getSection`, but MediaWiki API requires its index when updating it, and Wikimate passes that requirement right on to the user. This means in most scripts one has to add extra code to compute the index number from the section name.

I would like to see this functionality added to `setText` (and thus `setSection`) so that user code can be simplified.

Xymph:
See #40. The new private method allows for easy and transparent processing of `$section` in `setText`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Changelog
 
+### Version 0.12.0
+
+* Supports a section name (in addition to an index) in `WikiPage::setText()` and `WikiPage::setSection()`
+* Bug fix: prevent PHP Notices in several methods
+
 ### Version 0.11.1
 
 * Bug fix: handle unknown section parameter correctly in `WikiPage::getSection()`
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -605,12 +605,19 @@ class WikiPage {
 	
 	/**
 	 * Sets the text in the page.  Updates the starttimestamp to the timestamp
-	 * after the page edit (if the edit is successful)
+	 * after the page edit (if the edit is successful).
+	 *
+	 * Section can be the following:
+	 * - section name (string:"History")
+	 * - section index (int:3)
+	 * - a new section (string:"new")
+	 * - the whole page (null)
 	 *
 	 * @param   string   $text     The article text
-	 * @param   string   $section  The section to edit (null for whole page)
+	 * @param   string   $section  The section to edit (whole page by default)
 	 * @param   boolean  $minor    True for minor edit
-	 * @param   string   $summary  Summary text
+	 * @param   string   $summary  Summary text, and section header in case
+	 *                             of new section
 	 * @return  boolean            True if page was edited successfully
 	 */
 	function setText( $text, $section = null, $minor = false, $summary = null ) {
@@ -682,9 +689,16 @@ class WikiPage {
 	 * Essentially an alias of WikiPage:setText()
 	 * with the summary and minor parameters switched.
 	 *
+	 * Section can be the following:
+	 * - section name (string:"History")
+	 * - section index (int:3)
+	 * - a new section (string:"new")
+	 * - the whole page (null)
+	 *
 	 * @param   string   $text     The text of the section
-	 * @param   mixed    $section  Section index, new by default
-	 * @param   string   $summary  Summary text
+	 * @param   mixed    $section  The section to edit (intro by default)
+	 * @param   string   $summary  Summary text, and section header in case
+	 *                             of new section
 	 * @param   boolean  $minor    True for minor edit
 	 * @return  boolean            True if the section was saved
 	 */
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -1,8 +1,8 @@
 <?php
 /// =============================================================================
 /// Wikimate is a wrapper for the MediaWiki API that aims to be very easy to use.
-/// 
-/// @version    0.11.1
+///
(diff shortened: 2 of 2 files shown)


waldyrious:
I thought I had mentioned it, but can't find a comment to taht effect -- I think it's not necessary to update versions for each of these features until a version is actually released. I.e., bug fixes and breaking changes don't need to automatically produce a new version, they can be accumulated and the version increase will correspond to the largest of the changes included in it. I know it's up to me to make that happen, but on the other hand I don't think it makes sense to make several releases in quick succession. Let me know it you disagree, though.

Xymph:
Good that you bring this up, I'm limping on two thoughts. One is that I agree that it isn't worthwhile to increase versions except at releases, and incrementing/changelogging at every tiny brach merge would just clutter things up. That's why I didn't increment to 0.11.2 in #43.

The other is that every merge of a branch back into master is a release-of-sorts, and it should be clear what is merged from the changelog so that log and code correlate. That's why I mentioned the #43 fix in 1e431bd as yet, even though it belonged to another branch. I don't think it's a good idea to hold off updating Changelog until an actual Release, because then the maintainer/collaborator has to traverse back over possibly-umpteen merges to cull summaries, and forgetting/misinterpreting an entry is not unlikely. Though possibly existing tooling can assist with that.

And one might even distinguish before and after the merge via the version number. After all, at any point a user can clone a new fork from the master -- would be useful to know what's in it and from where it originates.

But I'm sure many smarter minds than mine :) have mulled over this, so I looked around a bit. GitVersion [indeed](https://gitversion.readthedocs.io/en/latest/#assumptionsrules) [doesn't](https://gitversion.readthedocs.io/en/latest/more-info/version-increments/) increment all the time. And here's an [opposing opinion](https://gist.github.com/jashkenas/cbd2b088e20279ae2c8e) to the SemVer approach (neither of which I fully digested). But I can't find much in the way of what people consider best practices on Git(Hub).

So I guess it's up to us what we'd like to do here, as this isn't a really full-fledged project yet and we can still shape its direction. What would you propose?

waldyrious:
> The other is that every merge of a branch back into master is a release-of-sorts

Right -- that makes sense in projects following a git branch management like git-flow, where there's a development branch and a master branch, which in fact ends up working as a release candidate branch. There are tons of conflicting opinions on this, but the de-facto status is that most small git projects (on github especially, but not exclusively) simply use master as the development branch, and bugfix/feature branches merged into it via PRs; pushing directly to master is avoided for multi-user projects, except for minor tweaks like [this one](https://github.com/hamstar/Wikimate/commit/3579e53), although this is usually only enforced by convention.

(Github actually does a great job at showing the provenance and context of each commit (even non-merge ones) by providing the PR number they're associated with, e.g. 26b07a1. But that's just a neat extra, we can't rely on that alone.)

Anyway, for a project the size of Wikimate, I think it's pretty reasonable to use this simplified workflow, and make releases by making a commit changing the version and tagging it (and creating a release on github & relevant package managers, although these steps are not related to git itself).

This does NOT mean that the changelog will only be updated on releases: several projects actually implement policy/convention of always updating the changelog for each change (except cleanup, minor documentation fixes, etc), adding the reference of the commit/PR the change relates to. For example, see https://github.com/mwclient/mwclient/blob/master/RELEASE-NOTES.md (or, for a larger project, https://github.com/JuliaLang/julia/blob/master/HISTORY.md). We can certainly do that, which saves us the hassle of sifting through the commits/diff since the last release in order to build the changelog of a new release :)

> Though possibly existing tooling can assist with that.

Yes, actually :) for example, [conventional-changelog and others](https://github.com/conventional-changelog/conventional-changelog#related). But I never used them, and am not sure how it detects which changes/commits/PRs to actually include in the changelog (i.e. how it could omit minor ones that would add more noise than signal, or combine a series of changes that are related, etc). So I guess for now it's best to stick to a manual approach as I described above.

What do you say?

Xymph:
I completely agree, a manual approach is best for a project of this size, and I do like the method of 
adding changelog entries with issue numbers as PRs are processed. Not having to collect them at release time anymore was exactly my point.

So what do we do with CHANGELOG.md as it currently stands, with intermediate version numbers since 0.10.0 that we would like to prune now? And what would the heading of the in-development log be called?

Would it be better to create a separate issue/PR for straightening out that file (and make one more commit here to undo the version/changelog stuff of 1e431bd), or can we use this PR to restructure things?

waldyrious:
> So what do we do with CHANGELOG.md as it currently stands, with intermediate version numbers since 0.10.0 that we would like to prune now?

We can do as you suggested, fix it up with a separate PR.

> And what would the heading of the in-development log be called?

"Next release" sounds fine to me, but anything along those lines would work. We can't give it a specific version number because that depends on the types of changes that end up going into it.

> Would it be better to create a separate issue/PR for straightening out that file (and make one more commit here to undo the version/changelog stuff of 1e431bd), or can we use this PR to restructure things? I would prefer the former, as it's cleaner to first wrap up this ticket and #38 before molding Changelog into the desired structure/content in preparation for the next release.

Yep, agreed.

waldyrious:
The changes in this PR look good. I'll merge now :)

PR #46, opened 2016-10-07
Declare visibility on all methods, move opening braces (fixes #44)

### LINKED ISSUE #44: WikiPage methods: visibility should be declared
[Per coding guidelines](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#43-methods). Same for the Wikimate constructor (all its other methods already have declarations).

Also, class & method opening braces could go on separate lines.

Xymph:
A straightforward commit for #44, little more to say about it.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -12,7 +12,8 @@
  * @author  Robert McLeod
  * @since   December 2010
  */
-class Wikimate {
+class Wikimate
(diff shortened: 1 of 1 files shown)


waldyrious:
Straightforward improvement, thanks!

PR #49, opened 2016-10-09
Collect changes since last release under current version (fixes #48)

### LINKED ISSUE #48: Rework Changelog and recent version numbering
See discussion in #45:
- Collect future improvements/fixes in Changelog following from each issue under a heading like "Next release" or "Current development" or somesuch.
- Append issue numbers to all Changelog entries where applicable.
- Assign a new version in Wikimate.php and Changelog only at release time. Therefore, take out some current intermediate versions?

Xymph:
Kept v0.12.0 (in Wikimate.php too) as it would be confusing to go back now.

@waldyrious Could you comment on #38?

Btw, to test-render .md files, [this site](https://jbt.github.io/markdown-editor/) is quite useful. :+1:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,17 +2,18 @@
 
 ### Version 0.12.0
 
-* Supports a section name (in addition to an index) in `WikiPage::setText()` and `WikiPage::setSection()`
-* Improved section processing in `WikiPage::getText()`
-* Supports optional domain at authentication
-* Bug fix: prevent PHP Notices in several methods
-* Bug fix: handle unknown section parameter correctly in `WikiPage::getSection()`
-* Bug fix: pass return value in `WikiPage::setSection()`
-* Bug fix: correct call to `Wikimate::debugRequestsConfig()`
+* Supports a section name (in addition to an index) in `WikiPage::setText()` and `WikiPage::setSection()` (#45)
+* Improved section processing in `WikiPage::getText()` (#33, #35)
+* Supports optional domain at authentication (#28)
+* Restructured and improved documentation (#32, #34, #47)
+* Bug fix: prevent PHP Notices in several methods (#43)
+* Bug fix: handle unknown section parameter correctly in `WikiPage::getSection()` (#41)
+* Bug fix: pass return value in `WikiPage::setSection()` (#30)
+* Bug fix: correct call to `Wikimate::debugRequestsConfig()` (#30)
 
 ### Version 0.10.0
 
-* Switched to using the *Requests* library instead of Curl
+* Switched to using the *Requests* library instead of Curl (#25)
 
 ### Version 0.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,19 +1,19 @@
 ## Changelog
 
-### Version 0.12.0
+### Version 0.11.0
 
-* Supports a section name (in addition to an index) in `WikiPage::setText()` and `WikiPage::setSection()` (#45)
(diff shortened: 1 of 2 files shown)


Xymph:
I chose PR numbers instead of issue numbers for the change entries, because the latter don't always exist for some changes (like domain support). Prior to v0.10.0 there seems to be little correlation between issues/PRs on the one hand and changelog entries on the other, probably because hamstar committed everything directly to master. Therefore I didn't add numbers to older entries.

waldyrious:
If your commit will include edits to a single file, you can actually edit it directly on github, which provides you with a preview tab you can use before submitting the commit. In fact, you can even use the preview feature of github as a markdown previewer only, then cancel the change on the web UI and submit the change via the command line, if you prefer it that way. The advantage of this approach is that you're guaranteed to get the exact rendering that GitHub will use after the commit is merged.

waldyrious:
Looks good, but I have to comments:

- I'd rather use full markdown links, like here: https://raw.githubusercontent.com/JuliaLang/julia/master/NEWS.md, because the auto-linking is a github-specific feature and won't work if the file is being viewed elsewhere (i.e. in a local copy of the repo)
- Why jump from v0.10.0 to v0.12.0? Shouldn't the next version be 0.11.0?

Xymph:
Yes, previewing .md in the GUI is also an option, but indeed I prefer to use CLI git consistently now that I got the hang of it. :smile: 

By "full markdown links" you mean the square brackets around the numbers, and the list at the end "generated by NEWS-update.jl" with the full URLs?  But I don't have Julia (and no intention of installing it) so can't generate that list using their script. Got an alternative? And where is that markdown feature described?

v0.12.0 is already public (if only a few days), I felt it would be a bit weird to go back. Also, jumps aren't so rare, this very repo went from 0.5 to 0.10.0. But if you think 0.11.0 is better, that's fine with me too.

waldyrious:
> By "full markdown links" you mean the square brackets around the numbers, and the list at the end "generated by NEWS-update.jl" with the full URLs? 

Yes.

>  But I don't have Julia (and no intention of installing it) so can't generate that list using their script. Got an alternative?

Well, that script was created after the fact. The inclusion of the links was done [manually at first](https://github.com/JuliaLang/julia/commit/a0aa786), but for such a large project, it makes sense that they chose to automate it (also to remove errors of people forgetting to add the link or copy-pasting an existing link and forgetting to change it). For our case, that shouldn't be an issue :) that's how [this](https://github.com/mwclient/mwclient/commit/b82984d) is [done](https://github.com/mwclient/mwclient/commit/aa490ba) in [mwclient](https://github.com/mwclient/mwclient/commit/dffd7ba) for instance.

> And where is that markdown feature described?

Reference-style links have been present since the [very first version](https://daringfireball.net/projects/markdown/syntax#link) of markdown :)

> v0.12.0 is already public (if only a few days), I felt it would be a bit weird to go back. Also, jumps aren't so rare, this very repo went from 0.5 to 0.10.0. But if you think 0.11.0 is better, that's fine with me too.

If we're treating the master branch as a development branch, changes to it aren't considered a release. I mean, there isn't even a way to download 0.12.0, not even by checking out a git tag (since no such tag was created). So to "get v0.12.0" one would have to find the exact commit that introduced that version number, and do a checkout of it (ending up in a git detached head state). So I don't think it's problematic to discard 0.12.0 as an actual Wikimate version.

Xymph:
Okay, I manually added the reference list, easy enough for this small number of entries.

waldyrious:
Excellent work! 👍

PR #50, opened 2016-10-09
Improve regex for section header at end of page

Xymph:
I discovered that [this rationale](https://github.com/hamstar/Wikimate/pull/33#issue-180529597) is incomplete: getting the page text will not include a newline after that last line (talk about unexpected...), therefore if a section header is the last line on the page, the regex will fail to match that header.

This commit fixes it, by accepting (but not capturing) a newline _or_ the end of the text.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -481,7 +481,7 @@ class WikiPage
 			unset( $page );
 			
 			// Now we need to get the section headers, if any
-			preg_match_all( '/(={1,6}).*?\1 *\n/', $this->text, $matches );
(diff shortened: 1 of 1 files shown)


waldyrious:
Looks good, although this should be a fairly uncommon case :)

Xymph:
Maybe, but I ran into it while preparing for a bot run, so it had to be addressed.

waldyrious:
> it had to be addressed.

Definitely, even if you hadn't come across it :)

Issue #51, opened 2016-10-17
getSection: not include heading fails for empty section


Xymph:
If `$includeHeading` is false, the ==Header== should be chopped from the returned text, but this fails if the section has an empty body: then the header is returned instead of the empty string I'd expect.

PR #52, opened 2016-10-17
Fix returning empty section w/o header

Xymph:
See #51. The `trim` on a section without body leaves nothing but the header without trailing newline. Thus `strpos` fails and `substr` returns the original section text including header. The `trim` simply isn't needed to find the first newline, as the `preg_match_all` pattern already insures the section text begins with at least one `=` (so a `ltrim` is not needed either).


waldyrious:
Nice catch :)

PR #55, opened 2016-10-23
Modify getSection to optionally include subsections (fixes #53)

### LINKED ISSUE #53: Add subsections optionally to getSection result
`setSection/setText` write the specified section contents with the provided text as is. MediaWiki API uses that text to replace the entire section, including any subsections. But WikiPage discards any nesting structure in the page, and `getSection` only returns the initial portion of a section, not its subsections. That's convenient when the initial section is all one wants to change, but there's no convenient way to write that back without discarding its subsection(s).

I don't think `setSection/setText` should be changed as they follow the API mechanism. And sometimes one _does_ want to discard all subsections (as has been the case in a previous bot script I wrote) or the user already knows to be working on a subsection that has no further subsections.

Instead I propose to add a new method `getSubsections($section)` so that the user can do something like this:
```
$oldsection = getSection(1); // or 'name'
$newsection = transform($oldsection);
$newsection .= getSubsections(1); // or 'name'
setSection($newsection);
```
This also will require adding a field to the two `$this->sections` arrays to keep track of the nesting level or depth, i.e. the number of `=`'s of each header.

Comments (@waldyrious)?

Xymph:
See #53. Tracking section depth is needed to support old and new behavior of getSection. And a user of `getSectionOffsets` may find another use for it too.

In `getSection` if `includeSubsections` is true, the new `foreach` needs to loop over any sections following the requested one and add their lengths to the initial one. It can use either the `byIndex` or `byName` array, that doesn't matter, and first iterates until the same offset as the requested section, then adds those lengths until an entry is no longer a subsection.

Yes, already had this algorithm roughly in my mind even before creating the issue. :sunglasses:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,11 @@
 
 ### Version 0.11.0
 
+* `WikiPage::getSection()` now includes subsections by default; disabling the new $includeSubsections option reverts to the old behavior of returning only the text until the first subsection ([#55])
 * Supports a section name (in addition to an index) in `WikiPage::setText()` and `WikiPage::setSection()` ([#45])
-* Improved section processing in `WikiPage::getText()` ([#33], [#37], [#50])
+* Improves section processing in `WikiPage::getText()` ([#33], [#37], [#50])
 * Supports optional domain at authentication ([#28])
-* Restructured and improved documentation ([#32], [#34], [#47], [#49])
+* Restructures and improves documentation ([#32], [#34], [#47], [#49])
 * Bug fix: return empty section without header in `WikiPage::getSection()` ([#52])
 * Bug fix: prevent PHP Notices in several methods ([#43])
 * Bug fix: handle unknown section parameter correctly in `WikiPage::getSection()` ([#41])
@@ -49,4 +50,5 @@
 [#49]: https://github.com/hamstar/Wikimate/pull/49
 [#50]: https://github.com/hamstar/Wikimate/pull/50
 [#52]: https://github.com/hamstar/Wikimate/pull/52
+[#55]: https://github.com/hamstar/Wikimate/pull/55
 
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -556,17 +556,17 @@ class WikiPage
 	}
 	
 	/**
-	 * Returns the section requested, with its subsections, if any.
+	 * Returns the requested section, with its subsections, if any.
 	 *
 	 * Section can be the following:
 	 * - section name (string:"History")
 	 * - section index (int:3)
 	 *
 	 * @param   mixed    $section             The section to get
-	 * @param   boolean  $includeHeading      False to get section text only, true
-	 *                                        to include heading too
-	 * @param   boolean  $includeSubsections  False to get section text only, true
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,10 +2,11 @@
 
 ### Version 0.11.0
 
+* `WikiPage::getSection()` now includes subsections by default; disabling the new $includeSubsections option reverts to the old behavior of returning only the text until the first subsection ([#55])
 * Supports a section name (in addition to an index) in `WikiPage::setText()` and `WikiPage::setSection()` ([#45])
-* Improved section processing in `WikiPage::getText()` ([#33], [#37], [#50])
+* Improves section processing in `WikiPage::getText()` ([#33], [#37], [#50])
```
waldyrious: Honestly, I prefer the old phrasing.

Wikimate.php
```diff
@@ -552,18 +556,21 @@ public function getText( $refresh = false )
 	}
 	
 	/**
-	 * Returns the section requested.
+	 * Returns the section requested, with its subsections, if any.
```
waldyrious: maybe change "section requested" --> "requested section"?

waldyrious:
Very nice :D Looks good to me, I left only a few inline comments.

CHANGELOG.md
Xymph: We're mixing present and past tense there, and imperative present in the bug fix entries. Mixing tenses is generally a bad idea. I used imperative due to [this recommendation](http://stackoverflow.com/a/3580764), but there are [opposing](http://stackoverflow.com/a/8059167) [opinions](http://softwareengineering.stackexchange.com/questions/157590/why-is-it-preferred-to-write-a-commit-message-in-present-tense-imperative-mood) too. For some entries, like this new PR, present tense ("now includes", "reverts") seems to work best, hence I made the others consistent. For some older entries, past tense ("Supported optional domain") wouldn't work, as that sounds like the project ceased to support it.

I'm okay with past tense (used that in most of my projects), but then some entries will need further rewriting. While we're at it, the number of 0.11.0 entries has grown somewhat large, so maybe we can group and prefix them according to [A good change log sticks to these principles](http://keepachangelog.com/en/0.3.0/)?

Xymph:
Thanks, I've applied your other suggestions, only the change log needs more discussion/work as outlined above.

CHANGELOG.md
waldyrious: I agree on the consistency. My preference would be to the imperative (which the old versions of these lines could be read as, if interpreted in the passive voice) over either past or present tense (I.e. a non-personal description of the change itself, rather than of what the change author intended). The first opposing opinion you linked to argues against present tense, but makes no mention of imperative, and the second one also makes no arguments against imperative. What do you think?

As for following the Keep A Changelog format, I'm all for it 😄, but let's do that in a separate PR.

Xymph:
Added a small boy scout commit. That's all for this PR from me. :)

waldyrious:
Great. I'll merge and will try to find a bit of time to edit the changelog, but won't promise anything -- over-commitment and spreading myself thin is kind of my specialty :/

PR #57, opened 2016-11-01
Handle invalid page title correctly (fixes #56)

### LINKED ISSUE #56: Invalid page title echoes error
When specifying an invalid page title (e.g. with a trailing ":") to WikiPage, it echoes an error message. It should use the error property instead.

Also, when getText loads such an invalid title, it should not continue processing it.

Various minor code and comment improvements will also be included in this PR.

Xymph:
See #56. Initial commit aborts getText processing up invalid title, and puts the message in the error property.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Improves section processing in `WikiPage::getText()` ([#33], [#37], [#50])
 * Supports optional domain at authentication ([#28])
 * Restructures and improves documentation ([#32], [#34], [#47], [#49])
+* Bug fix: handle invalid page title correctly ([#57])
 * Bug fix: return empty section without header in `WikiPage::getSection()` ([#52])
 * Bug fix: prevent PHP Notices in several methods ([#43])
 * Bug fix: handle unknown section parameter correctly in `WikiPage::getSection()` ([#41])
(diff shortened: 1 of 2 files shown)


waldyrious:
Simple and straightforward 👍 LGTM! Feel free to merge whenever you're ready :)

Xymph:
Added a handful of "boy scout" changes :smile: as well that I wanted to get out of the way before the release tag is made (after which a whole new class will be added...). Please remember the Changelog re-org. :wink: 

How should the merge be done, using which option from the "Merge PR" menu?

Xymph:
Hmm, I wanted to rebase/fixup 3d72a32 into eff91ac, but ran into a problem (Automatic cherry-pick failed). Attempting to fix it probably made it worse, and git also threw me out of the local 56-fix-invalid-page branch. So unless you have a suggestion how to untangle this, I'll just leave the remote commits alone.

Xymph:
More overdue housekeeping applied. Now it really is enough. :wink:

waldyrious:
I'll leave this one for you to merge so you can test the merge button :) In this case, since we want to preserve the individual commits, the appropriate option is "merge" ("rebase and merge" should be similar, but I guess it only makes a difference if the PR's branch was created off a branch other than master). The "squash" option would not create a merge commit, but instead condense all of the PR's commits into a single one and add it to the master branch. That's only desirable when the commit separation was excessive (e.g. if an additional commit was added to fix build errors or a typo in the previous commit, rather than to provide semantically independent changes). Let me know if anything isn't clear.

PR #63, opened 2016-11-09
Handle MediaWiki API and other errors consistently (fixes #59)

### LINKED ISSUE #59: Inconsistent value storing in $this->error
When `$this->error` is filled from an API response, it's done in statements like this:
```
// Check for errors
if ( isset( $r['error'] ) ) {
    $this->error = $r; // Set the error if there was one
} else {
    $this->error = null; // Reset the error status
}
```
So `$this->error` becomes an array containing one key 'error' with an array as value, containing one or more key/value pairs with MediaWiki's errors. The interesting values are thus nested one level, to little purpose IMHO.

But other code, both existing and recently added by me, simply add an array element to `$this->error` directly, e.g. `$this->error['login']` and `$this->error['section']`.

While that's inconsistent, I'm not sure which way to make this consistent. Getting rid of the nesting level in API errors (i.e. using `$this->error = $r['error'];` above) might break existing code out there. Inserting the extra level at the other assignments seems a bit silly, too.

What do you think, @waldyrious?

Xymph:
See #59. Commit 7fc1c0f ensures that MW API errors in `$r['error']` are copied to `$this->error` in all cases where it may be returned. In `getText`, line 463 also ceases further processing in case of an API error - previously it would still parse the response and attempt to extract sections if the `invalid` flag wasn't set in the response.

The `setText` handling of API responses became a little more complex than I first anticipated. Earlier in the method it checks `$r['edit']['result']` for `Success` but that implies a different result is also possible, instead of an `$r['error']` array. The [MW edit page](https://www.mediawiki.org/wiki/API:Edit#CAPTCHAs_and_extension_errors) describes the alternative, `Failure`. In order to keep WikiPage simple and easy, I didn't handle such situations in detail, the `isset` merely avoids a PHP notice if `$r['error']` is undefined because of the Failure result, and that result is passed in a custom message.

In the `delete` method all this isn't needed (line 818) because there's no `'result'` element with different possible values in the `'delete'` response.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Since v0.10.0 this project adheres to [Semantic Versioning](http://semver.org/) 
 
 * Updated `WikiPage::getSection()` to include subsections by default; disabling the new `$includeSubsections` option reverts to the old behavior of returning only the text until the first subsection ([#55])
 * Improved section processing in `WikiPage::getText()` ([#33], [#37], [#50])
+* Ensured that MediaWiki API error responses appear directly in `WikiPage::$error` rather than a nested 'error' array ([#63]) -- this may require changes in your application's error handling
 * Restructured and improved documentation ([#32], [#34], [#47], [#49], [#61])
 
 #### Fixed
@@ -67,4 +68,5 @@ Since v0.10.0 this project adheres to [Semantic Versioning](http://semver.org/) 
 [#55]: https://github.com/hamstar/Wikimate/pull/55
 [#57]: https://github.com/hamstar/Wikimate/pull/57
 [#61]: https://github.com/hamstar/Wikimate/pull/61
+[#63]: https://github.com/hamstar/Wikimate/pull/63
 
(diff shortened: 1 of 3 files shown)


Xymph:
In 3aac5fc the `error` property is initialized to an empty array prior to setting a custom message. While not strictly necessary because it starts out at `null` at object creation, if a user calls a new method after the preceding one gave an error, it might mix different key/value pairs, and the init prevents that.

Also, line 858 standardizes the custom key/value pairs within WikiPage so the user would need to check only one key. Lastly, I made quoting a bit more consistent.

Xymph:
4143ef5 also gets rid of a dangling empty `else` branch.

Xymph:
That should be it. I only would like to take out the accidental commit (-am can be dangerous...) and its revert. Can that be done via rebase -i and fixups for 268f7f5 and 614268c?

waldyrious:
> I only would like to take out the accidental commit (-am can be dangerous...) and its revert. Can that be done via rebase -i and fixups for 268f7f5 and 614268c?

Yes, I do that all the time :) If you want, you can do a `git branch "<backup_ref>"` before the rebase, so that if the `rebase -i` doesn't turn out like you wanted, you can simply reset it to where it was before by doing `git branch -f <rebased_branch> <backup_ref>`, and try again.

waldyrious:
By the way, stellar work documenting these changes! 👌

Xymph:
Thank you.
The rebase/fixups went fine, time to merge. And fix the commit #'s above, as they all changed and no longer document as well. :wink:

PR #69, opened 2016-11-16
Add new class WikiFile (fixes #58)

### LINKED ISSUE #58: New class WikiFile
I need the ability to get information about files, and to upload them. I've started writing a new `WikiFile` class for this, still need to write the upload (and download, while I'm at it) methods. But I'd like to commit it after the changelog re-org and release tag are done, in order not to mingle branches/commits too much. So this is your cue, @waldyrious. :sunglasses:

Xymph:
See #58. The first commit adds the class with all magic methods, getters, and download/upload methods. The second one integrates it with Wikimate and provides the underlying session actions for download/upload.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -858,3 +858,501 @@ class WikiPage
 		return -1;
 	}
 }
+
+
+/**
+ * Models a wiki file that can have its properties retrieved and
+ * its contents downloaded and uploaded.
+ * All properties pertain to the current revision of the file.
+ *
+ * @author  Robert McLeod & Frans P. de Vries
+ * @since   October 2016
+ */
+class WikiFile
+{
+	protected $filename  = null;
+	protected $wikimate  = null;
+	protected $exists    = false;
+	protected $invalid   = false;
+	protected $error     = null;
+	protected $edittoken = null;
+	protected $info      = null;
+
+	/*
+	 *
+	 * Magic methods
+	 *
+	 */
+
+	/**
+	 * Constructs a WikiFile object from the filename given
+	 * and adds a Wikimate object.
+	 *
+	 * @param  string    $filename  Name of the wiki file
+	 * @param  Wikimate  $wikimate  Wikimate object
+	 */
+	public function __construct($filename, $wikimate)
+	{
+		$this->wikimate = $wikimate;
+		$this->filename = $filename;
+		$this->info     = $this->getInfo(true);
+
+		if ($this->invalid) {
+			$this->error['file'] = 'Invalid filename - cannot create WikiFile';
+		}
+	}
+
+	/**
+	 * Forget all object properties.
+	 *
+	 * @return  <type>  Destructor
+	 */
+	public function __destruct()
+	{
+		$this->filename   = null;
+		$this->wikimate   = null;
+		$this->exists     = false;
+		$this->invalid    = false;
+		$this->error      = null;
+		$this->edittoken  = null;
+		$this->info       = null;
+		return null;
+	}
+
+	/**
+	 * Returns the file existance status.
+	 *
+	 * @return  boolean  True if file exists
+	 */
+	public function exists()
+	{
+		return $this->exists;
+	}
+
+	/**
+	 * Alias of self::__destruct().
+	 */
+	public function destroy()
+	{
+		$this->__destruct();
+	}
+
+	/*
+	 *
+	 * File meta methods
+	 *
+	 */
+
+	/**
+	 * Returns an error if there is one, null shows no error.
+	 *
+	 * @return  mixed  Null for no errors, or an error array
+	 */
+	public function getError()
+	{
+		return $this->error;
+	}
+
+	/**
+	 * Returns the name of this file.
+	 *
+	 * @return  string  The name of this file
+	 */
+	public function getFilename()
+	{
+		return $this->filename;
+	}
+
+	/*
+	 *
+	 * Getter methods
+	 *
+	 */
+
+	/**
+	 * Gets the information of the file. If refresh is true,
+	 * then this method will query the wiki API again for the file details.
+	 *
+	 * @param   boolean  $refresh  True to query the wiki API again
+	 * @return  mixed              The info of the file (array), or null if error
+	 */
+	public function getInfo($refresh = false)
+	{
+		if ($refresh) { // We want to query the API
+
+			$data = array(
+				'titles' => 'File:' . $this->filename,
+				'prop' => 'info|imageinfo',
+				'iiprop' => 'bitdepth|canonicaltitle|comment|parsedcomment|'
+				          . 'commonmetadata|metadata|extmetadata|mediatype|'
+				          . 'mime|thumbmime|sha1|size|timestamp|url|user|userid',
+				'intoken' => 'edit',
+			);
+
+			$r = $this->wikimate->query($data); // Run the query
+
+			// Check for errors
+			if (isset($r['error'])) {
+				$this->error = $r['error']; // Set the error if there was one
+				return null;
+			} else {
+				$this->error = null; // Reset the error status
+			}
+
+			// Get the page (there should only be one)
+			$page = array_pop($r['query']['pages']);
+			unset($r, $data);
+
+			// Abort if invalid file title
+			if (isset($page['invalid'])) {
+				$this->invalid = true;
+				return null;
+			}
+
+			$this->edittoken = $page['edittoken'];
+
+			if (!isset($page['missing'])) {
+				// Update the existance if the file is there
+				$this->exists = true;
+				// Put the content into info
+				$this->info   = $page['imageinfo'][0];
+			}
+			unset($page);
+		}
+
+		return $this->info; // Return the info in any case
+	}
+
+	/**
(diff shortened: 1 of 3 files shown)


Wikimate.php
```diff
@@ -858,3 +858,501 @@ private function findSection($section)
...
+	{
+		return $this->error;
+	}
+
+	/**
+	 * Returns the name of this file.
+	 *
+	 * @return  string  The name of this file
+	 */
+	public function getFilename()
```
waldyrious: Shouldn't this be under the "getter methods" below?

Wikimate.php
```diff
@@ -858,3 +858,501 @@ private function findSection($section)
...
+		$this->exists     = false;
+		$this->invalid    = false;
+		$this->error      = null;
+		$this->edittoken  = null;
+		$this->info       = null;
+		return null;
+	}
+
+	/**
+	 * Returns the file existance status.
```
waldyrious: typo: existence

Wikimate.php
```diff
@@ -858,3 +858,501 @@ private function findSection($section)
...
+	 *
+	 * @return  string  The canonical title of this file
+	 */
+	public function getCanonicalTitle()
+	{
+		return $this->info['canonicaltitle'];
+	}
+
+	/**
+	 * Returns the comment of this file.
```
waldyrious: Maybe we could clarify what "comment" means in this context.

Wikimate.php
```diff
@@ -858,3 +858,501 @@ private function findSection($section)
...
+	 *
+	 * @return  string  The comment of this file
+	 */
+	public function getComment()
+	{
+		return $this->info['comment'];
+	}
+
+	/**
+	 * Returns the common metadata of this file.
```
waldyrious: It would be nice to clarify what common (as opposed to non-common) metadata refers to.

Wikimate.php
```diff
@@ -858,3 +858,501 @@ private function findSection($section)
...
+	{
+		return $this->info['descriptionurl'];
+	}
+
+	/**
+	 * Returns the extended metadata of this file.
+	 *
+	 * @return  array  The extended metadata of this file
+	 */
+	public function getExtendedMetadata()
```
waldyrious: This should probably come right after the commonmetadata method.

Wikimate.php
```diff
@@ -858,3 +858,501 @@ private function findSection($section)
...
+	{
+		return $this->info['extmetadata'];
+	}
+
+	/**
+	 * Returns the height of this file.
+	 *
+	 * @return  int  The height of this file
+	 */
+	public function getHeight()
```
waldyrious: Are these methods sorted alphabetically? I think it would be nice to sort them in semantically meaningful sections (e.g. height/width/aspect ratio, etc.)

waldyrious:
Gotta say, I'm really digging your use of atomic commits! Although I'd prefer if you added unrelated changes in separate PRs (even if you merge them right away, for being trivial changes).

As for the changes themselves, it looks good to me overall. I've added some inline comments, but they're mostly about making the code understandable to newcomers; functionally-wise, it seems ok to go.

Wikimate.php
Xymph: It currently isn't because of the alpha-sort.

Wikimate.php
Xymph: That is [MediaWiki API terminology](https://www.mediawiki.org/wiki/Special:MyLanguage/API:Imageinfo#Parameters), which isn't more descriptive about that either. All three Metadata properties depend entirely on the file type, so a user will always have to experiment before being able to apply them.

Wikimate.php
Xymph: Added the word 'edit' as per API description.

Wikimate.php
```diff
@@ -858,3 +858,501 @@ private function findSection($section)
...
+	/*
+	 *
+	 * File meta methods
+	 *
+	 */
+
+	/**
+	 * Returns an error if there is one, null shows no error.
+	 *
+	 * @return  mixed  Null for no errors, or an error array
```
Xymph: Changed everywhere.

Wikimate.php
Xymph: Like `getTitle` in WikiPage, this method returns the input parameter, not a file property (`getCanonicalTitle` does that), hence I'm keeping them in the "meta methods" section.

Wikimate.php
```diff
@@ -858,3 +858,501 @@ private function findSection($section)
...
+
+	/*
+	 *
+	 * Magic methods
+	 *
+	 */
+
+	/**
+	 * Constructs a WikiFile object from the filename given
+	 * and adds a Wikimate object.
```
Xymph: Just like in the WikiPage constructor, this is related to how Wikimate and the other two classes are strung together. `Wikimate::getFile` passes `$this` (PHP's variable for the current object, i.e. Wikimate) to the WikiFile constructor, so that WikiFile methods can then access the underlying Wikimate methods.

The user isn't supposed to called the WikiPage/WikiFile constructors directly, so I'm not sure how to clarify this, nor whether it's needed.

Wikimate.php
```diff
@@ -858,3 +858,501 @@ private function findSection($section)
...
+	protected $wikimate  = null;
+	protected $exists    = false;
+	protected $invalid   = false;
+	protected $error     = null;
+	protected $edittoken = null;
+	protected $info      = null;
+
+	/*
+	 *
+	 * Magic methods
```
Xymph: PHP's [OOP terminology](http://www.php.net/manual/en/language.oop5.magic.php) for constructor/destructor, toString, and so on.

Wikimate.php
```diff
@@ -858,3 +858,501 @@ private function findSection($section)
...
+			// Abort if invalid file title
+			if (isset($page['invalid'])) {
+				$this->invalid = true;
+				return null;
+			}
+
+			$this->edittoken = $page['edittoken'];
+
+			if (!isset($page['missing'])) {
+				// Update the existance if the file is there
```
Xymph: Fixed everywhere.

Wikimate.php
Xymph: Yes they are. Some can be grouped meaningfully indeed, but for others that is not so clear, hence I chose alphabetical as that at least makes it easier to look up a getter. See the MW API link above. If you'd want to group semantically, can you propose the new list?

Xymph:
Thanks for the quick review and kind words. As a general follow-up: WikiFile is based off WikiPage, so its structure and choice of words are derived from that (hence also the co-credit to @hamstar). So in some cases, if I changed something in WikiFile due to a review comment, I applied the same change in the other class(es).

As for the less-/un-related changes, you're the one pointing me to the boy scout principle. :wink: But I concede there's a vague line between how many such changes can hitchhike along, maybe this grew out to be too much. Either way, it saved me a little overhead (new branch, push, merge) to do it this way.

Wikimate.php
waldyrious: Well, if we're going to add documentation, we might as well make it clear. My issue is just with the "adds a Wikimate object" phrasing. Would "and associate with the passed Wikimate object" work?

Wikimate.php
waldyrious: fair enough. Speaks to my unfitness as maintainer of a PHP project :P

Wikimate.php
waldyrious: Ok, makes sense.

Wikimate.php
```diff
@@ -858,3 +932,501 @@ private function findSection($section)
...
+	 *
+	 * @return  string  The URL of this file
+	 */
+	public function getUrl()
+	{
+		return $this->info['url'];
+	}
+
+	/**
+	 * Returns the user of this file.
```
waldyrious: Would "the user who uploaded the file" be better?

Wikimate.php
waldyrious: I guess I'd only put width, height and aspect ratio together, since they're all closely related. But there isn't a clear obvious order for the rest of the properties, so I guess it's better to forgo the sorting / grouping of those three.

waldyrious:
> WikiFile is based off WikiPage, so its structure and choice of words are derived from that (hence also the co-credit to @hamstar).

Ah, that makes sense.

> So in some cases, if I changed something in WikiFile due to a review comment, I applied the same change in the other class(es).

Fair enough.

> As for the less-/un-related changes, you're the one pointing me to the boy scout principle. :wink: I concede there's a vague line between how many such changes can hitchhike along, maybe this grew out to be too much. Either way, it saved me a little overhead (new branch, push, merge) to do it this way.

Good point. I suppose my rule of thumb is to do general fixup in code that a change touches directly, rather than overall in the code. Since the classes in Wikimate live all in the same file, it's understandable that, even applying this principle, the cleanups may spill out to other parts of the file. It's no big deal anyway, that was a minor comment, really.

Wikimate.php
Xymph: Certainly.

Xymph:
Okay, a few more changes committed. Should I rebase/fixup them with the preceding one?

Then, is that a wrap? :)

waldyrious:
Yes to both questions :)

PR #71, opened 2016-11-21
Add uploadFromUrl and restructure other upload methods (fixes #70)

### LINKED ISSUE #70: Add upload from URL
PR #69 includes uploading local files, but MW API also allows [uploading from a URL](https://www.mediawiki.org/wiki/API:Upload#Uploading_from_URL) with just one different parameter. Even though I don't need it myself, I'd like to add it to WikiFile for completeness' sake. (The much more complex chunked uploading however, I'll leave for another enterprising contributor.)

Xymph:
See #70. Previously, `upload()` took $data and `uploadFile()` invoked `upload()`. However, `uploadFromUrl()` takes $url and otherwise needs exactly the same code as `upload()`. To allow sharing this common code, I had to restructure `upload()` a bit into a private method, `uploadCommon()` (maybe there's a better name for that?), and then its public use is made clearer by a new name, `uploadData()`.

The previous merge is just a few days old (so hasn't spread far yet) and still falls under "current development", so this API change seems acceptable to me, as it leads to consistent naming of the three public upload methods. For further consistency, I renamed `download()` to `downloadData()` as well.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -526,6 +526,7 @@ class WikiPage
 	public function getText($refresh = false)
 	{
 		if ($refresh) { // We want to query the API
+			// Specify relevant page properties to retrieve
 			$data = array(
 				'titles' => $this->title,
 				'prop' => 'info|revisions',
@@ -1053,7 +1054,7 @@ class WikiFile
 	public function getInfo($refresh = false)
 	{
 		if ($refresh) { // We want to query the API
-
+			// Specify all image properties to retrieve
 			$data = array(
 				'titles' => 'File:' . $this->filename,
 				'prop' => 'info|imageinfo',
@@ -1115,7 +1116,7 @@ class WikiFile
 	/**
 	 * Returns the bit depth of this file.
 	 *
-	 * @return  int  The bit depth of this file
+	 * @return  integer  The bit depth of this file
 	 */
 	public function getBitDepth()
 	{
@@ -1175,7 +1176,7 @@ class WikiFile
 	/**
 	 * Returns the height of this file.
 	 *
-	 * @return  int  The height of this file
+	 * @return  integer  The height of this file
 	 */
 	public function getHeight()
 	{
@@ -1235,7 +1236,7 @@ class WikiFile
 	/**
 	 * Returns the size of this file.
 	 *
-	 * @return  int  The size of this file
+	 * @return  integer  The size of this file
 	 */
 	public function getSize()
 	{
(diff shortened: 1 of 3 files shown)


Xymph:
Minor code change in addition to the restructuring: `is_null($text)` is a better way to test input parameters than `=== null`.

Xymph:
@waldyrious, would you be able to squeeze in a review one of these days? :wink:

Wikimate.php
```diff
@@ -1053,7 +1054,7 @@ public function getFilename()
 	public function getInfo($refresh = false)
 	{
 		if ($refresh) { // We want to query the API
-
+			// Specify all image properties to retrieve
```
waldyrious: Shouldn't either this line use "relevant ... properties", or line 529 use "all ... properties"? For consistency.

Wikimate.php
```diff
@@ -1406,6 +1404,29 @@ public function upload($data, $comment, $text = null, $overwrite = false)
 	}
 
 	/**
+	 * Uploads the given contents to the current file.
+	 * $text is only used for the article page of a new file, not an existing
```
waldyrious: I don't think a file's page is called "article page". I'd use just "page", or maybe "page contents".

Also, I'd split the line in the comma, since that helps the reading flow and conforms to [semantic linebreaking](http://rhodesmill.org/brandon/2012/one-sentence-per-line/).

waldyrious:
Apart from the minor comments I've left inline, this LGTM :)

waldyrious:
> @waldyrious, would you be able to squeeze in a review one of these days? :wink: 

Oh, crap, you were faster :P I had this page open for the last few days and didn't refresh it before starting the review. I'm saying this just to reassure you that I didn't review the PR now following your prompt, but indeed hadn't forgotten :)

Wikimate.php
Xymph: Well, it gets far more properties than for a page, and I meant it more in the sense of "list all those that are to be retrieved, and don't list those that won't be retrieved". :) But 'relevant' is fine too.

Either way, 'image' should be 'file', so thanks for highlighting this anyway.

Xymph:
Sorry. :wink: I guess all the quick review turnarounds in the past spoiled me a bit, but I thought I'd poll you after 5 days.
Thanks again, time to wrap up this one.

Wikimate.php
waldyrious: For future reference, #135 introduces the semantic line breaks recommendation explicitly into the project docs.

PR #73, opened 2016-11-27
minor wording change in documentation

waldyrious:
besides this being clearer, I believe the previous wording was not grammatically correct.


Xymph:
Ah, I suppose so. Too bad this came too late for the previous merge, but I know how that goes sometimes.

waldyrious:
No problem, I did give my approval to the other PR, so I take responsibility for the overlook.

PR #76, opened 2016-11-27
Add getHistory, getRevision, getArchivename, delete methods to WikiFile (fixes #72)

### LINKED ISSUE #72: Delete old revision of file
MW API has the ability to delete just an old(er) revision of a file, rather than a file description page with all its file revisions (like `WikiPage::delete` does). For this a `WikiFile::delete` method is needed, which in turn requires revision history information (specifically the archivename property), hence `getHistory`, `getRevision` and `getArchivename` methods.

It would even be possible to make all the getters work on a specified revision as well, but I'm still hesitating whether that helps or makes (some) things more complicated. More on that when I commit what I have so far, but I'd prefer to wrap up PR #71 first to avoid potential merge conflicts later.

Xymph:
See #72. Before, `getInfo()` only obtains the current revision, and I wanted to keep that the default as this suffices for most use cases, and it would be unnecessary strain on the MW API to always get all revisions. I struggled for a little while how to combine MW API's abilities regarding file revisions with Wikimate's caching approach (`getInfo()`'s `$refresh` parameter), but eventually settled on the following.

Enter `getHistory()` which by default just returns the same current revision but, if forced to refresh, can fetch the desired number of revisions or the max, and from the current revision backwards or using the timestamp(s). See the [API doc](https://www.mediawiki.org/wiki/API:Imageinfo#Parameters).

In the resulting history array, the keys are numbered from 0 upwards, and can subsequently be used to fetch one revision's properties array via `getRevision()` as well as an older revision's archive name via `getArchivename()`. If the history includes the current revision, which has no archive name, `getArchivename()` treats that as an error. If it doesn't include the current revision because of timestamp parameters, key 0 refers to the most recent revision in the selection.

The revision parameter can also be a timestamp, but this needs to be the exact one used in the properties. I tried to be more flexible and allow other formats, but translating them into ISO8601 in PHP code and then matching that against the cached property was problematic, also because of timezone settings/conversions. So I decided to keep it simple and clear by sticking to exact timestamp matches only.

The new `delete()` defines the File: page path itself now and optionally takes an archive name to delete an older revision, which affects (or doesn't affect) the `$exists` flag too.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -228,14 +228,53 @@ if ($file->exists()) die('Example image already exists');
 $result = uploadFromUrl('https://upload.wikimedia.org/wikipedia/en/a/a9/Example.jpg', 'Adopt Wiki example image');
 ```
 
-### Deleting...
+### History...
 
-If the account you're using has delete permissions, you can delete files as well via `delete()` on its description page:
+The revision history of a file can be obtained as an array with a properties array per revision:
 
 ```php
-$page = $wiki->getPage('File:Old-button.png');
+$file = $wiki->getFile('Often-changed-file.zip');
+// get only the current revision
+$history = $file->getHistory();
+// get the maximum number of revisions (500 for user accounts, 5000 for bot accounts)
+$history = $file->getHistory(true);
+// get the latest 10 revisions during 2015 (can use all MediaWiki timestamp formats)
+$history = $file->getHistory(true, 10, '2015-01-01 00:00:00', '2015-12-31 23:59:59');
+// iterate over revisions and print properties
+foreach ($history as $revision => $properties) {
+  echo "Revision $revision properties:\n";
+  print_r($properties);
+}
+```
+
+One specific revision can be requested by revision sequence number or by exact timestamp in the current history, as can its archive name:
+
+```php
+// get the latest 50 revisions
+$history = $file->getHistory(true, 50);
+// get the properties of the penultimate revision
+$revision = $file->getRevision(1);
+// get the archive name of the specific revision (must be ISO 8601 format)
+$archivename = $file->getArchivename('2016-11-22T33:44:55Z');
+```
+
+### Deleting...
+
+If the account you're using has delete permissions, you can delete files as well:
+
+```php
+$file = $wiki->getFile('Old-button.png');
 // returns true if the delete was successful
-$page->delete('The button was superseded by a new one');
+$file->delete('The button was superseded by a new one');
+```
+
+To delete a specific older revision of the file, the archive name is needed:
+
+```php
+$file = $wiki->getFile('Often-changed-file.zip');
+$history = $file->getHistory(true);
+$archivename = $file->getArchivename(3);
+$file->delete('This was an inadvertent release', $archivename);
 ```
 
 ### Other stuff
--- a/USAGE.md
+++ b/USAGE.md
@@ -233,7 +233,7 @@ $result = uploadFromUrl('https://upload.wikimedia.org/wikipedia/en/a/a9/Example.
 The revision history of a file can be obtained as an array with a properties array per revision:
 
 ```php
-$file = $wiki->getFile('Often-changed-file.zip');
+$file = $wiki->getFile('Frequently-changed-file.zip');
 // get only the current revision
 $history = $file->getHistory();
 // get the maximum number of revisions (500 for user accounts, 5000 for bot accounts)
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -952,6 +952,7 @@ class WikiFile
 	protected $error     = null;
 	protected $edittoken = null;
 	protected $info      = null;
+	protected $history   = null;
 
 	/*
 	 *
@@ -991,6 +992,7 @@ class WikiFile
 		$this->error      = null;
 		$this->edittoken  = null;
 		$this->info       = null;
+		$this->history    = null;
 		return null;
 	}
 
@@ -1049,9 +1051,10 @@ class WikiFile
 	 * then this method will query the wiki API again for the file details.
 	 *
 	 * @param   boolean  $refresh  True to query the wiki API again
+	 * @param   array    $history  An optional array of revision history parameters
 	 * @return  mixed              The info of the file (array), or null if error
 	 */
-	public function getInfo($refresh = false)
+	public function getInfo($refresh = false, $history = null)
 	{
 		if ($refresh) { // We want to query the API
 			// Specify relevant file properties to retrieve
@@ -1063,6 +1066,14 @@ class WikiFile
 				          . 'mime|thumbmime|sha1|size|timestamp|url|user|userid',
 				'intoken' => 'edit',
 			);
+			// Add optional history parameters
+			if (is_array($history)) {
+				foreach ($history as $key => $val) {
+					$data[$key] = $val;
+				}
+				// Retrieve archive name property as well
+				$data['iiprop'] .= '|archivename';
+			}
 
 			$r = $this->wikimate->query($data); // Run the query
 
@@ -1089,8 +1100,9 @@ class WikiFile
 			if (!isset($page['missing'])) {
 				// Update the existence if the file is there
 				$this->exists = true;
-				// Put the content into info
-				$this->info   = $page['imageinfo'][0];
+				// Put the content into info & history
+				$this->info    = $page['imageinfo'][0];
+				$this->history = $page['imageinfo'];
 			}
 			unset($page);
 		}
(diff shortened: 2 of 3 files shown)


waldyrious:
Heads up, I probably won't be able to review this until Thursday.

Xymph:
As hinted in #72, it is feasible to make all the earlier getters work on revisions too, something like:
```
get<Property>($revision = null):
if (isset($revision)) return info[<property>]
else do what getArchivename does
```
This generalizes the getters so that a user can do:
```
$sha1_curr = getSha1();
$history = getHistory(true);
$sha1_prev = getSha1(1);
```
instead of:
```
[...]
$sha1_prev = $history[1]['sha1'];
```
But with every generalization, usage becomes a bit more complex (although not so much here, as the default behavior remains the same), and (Doxygen) documentation becomes more elaborate.

What do you think, @waldyrious?

Xymph:
> Heads up, I probably won't be able to review this until Thursday.

No problem, take your time. Better get all this right on the first go. :)

Wikimate.php
```diff
@@ -1306,6 +1318,157 @@ public function getWidth()
...
+		return $this->history;
+	}
+
+	/**
+	 * Returns the properties of the specified file revision.
+	 *
+	 * Revision can be the following:
+	 * - revision timestamp (string:"2001-01-15T14:56:00Z")
+	 * - revision index (int:3)
+	 * The current revision has index 0.
```
waldyrious: Just to be explicit, it would be nice to mention that the revision index increases towards older revisions.

Wikimate.php
```diff
@@ -1306,6 +1318,157 @@ public function getWidth()
...
+			foreach ($this->history as $history) {
+				if ($history['timestamp'] == $revision) {
+					return $history;
+				}
+			}
+		}
+
+		// Return error message
+		$this->error = array();
+		$this->error['file'] = 'The revision is not found for this file';
```
waldyrious: I'd mention the actual revision parameter passed in the error message, and also change "is not found" --> "was not found".

Wikimate.php
```diff
@@ -1306,6 +1318,157 @@ public function getWidth()
...
+		return null;
+	}
+
+	/**
+	 * Returns the archive name of the specified file revision.
+	 *
+	 * Revision can be the following:
+	 * - revision timestamp (string:"2001-01-15T14:56:00Z")
```
waldyrious: is the colon a standard format for examples? If not, I'd use something like `(string, e.g. 2001-01-15T14:56:00Z")`, and maybe mention that it is in ISO 8601 format.

waldyrious:
I like the changes, but I'm a bit uneasy that `getHistory` only gets the latest revision by default. A sensible limit (e.g. 25 as IIRC is used in the page history in the mediawiki UI) would make more sense IMHO. I agree that for getInfo fetching only the last revision makes sense, but for getHistory this seems counter-intuitive. I also am not sure about getRevisions's behavior depending on a previous call of getHistory (if I got that right). I'm open to hear your rationale for these choices, of course.

Wikimate.php
Xymph: Examples were already [in that format](https://github.com/hamstar/Wikimate/blob/f0558f015484bce35ef32dd7fb075965d688c676/Wikimate.php#L487) before I came on board. I'm fine with changing it, but then I'll do it throughout the entire file - okay?

Wikimate.php
Xymph: Yes, and actually it is only the current revision if no timestamps were specified to `getHistory()`, so 'most recent' is more accurate.

Wikimate.php
```diff
@@ -1306,6 +1318,157 @@ public function getWidth()
...
+			$this->error = array();
+			$this->error['file'] = 'This revision contains no archive name';
+			return null;
+		}
+
+		return $info['archivename'];
+	}
+
+	/**
+	 * Delete the file, or an older revision of it.
```
Xymph: Agreed.

USAGE.md
```diff
@@ -228,14 +228,53 @@ if ($file->exists()) die('Example image already exists');
 $result = uploadFromUrl('https://upload.wikimedia.org/wikipedia/en/a/a9/Example.jpg', 'Adopt Wiki example image');
 ```
 
+### History...
+
+The revision history of a file can be obtained as an array with a properties array per revision:
+
+```php
+$file = $wiki->getFile('Often-changed-file.zip');
```
Xymph: Neither am I, but I should have written it that way to begin with.

Xymph:
Yes, `getRevision()` requires a preceding `getHistory()` call if getting anything but the most recent entry. Users dealing with revisions will already needed to know better what they're doing, so this mechanism seems acceptable to me.

My rationale was two-fold:
- Not retrieving more data than necessary in the initial API request, i.e. the default current revision instead of an arbitrarily larger number. I don't know what the default in the UI is (kinda hard to find a file with 25+ revs), but that seems a pretty large number already when -- in most use cases for WikiFile -- it's never needed.
- Make the caching behavior mimic that of WikiPage, so the initial `Wikimate::getFile()` makes one API request implicitly, and getting further/different data always requires an explicit `$refresh = true` to force another API request.

In order to ensure `getHistory()` behaves consistently even if not yet forced to refresh, it then follows that it should return the current revision, and not `null` or some such.

As noted in my initial post, I did struggle a bit with fitting revisions into the Wikimate/WikiFile model, and this seems a reasonably clean approach to it, but I'm open to entirely different approaches.

Wikimate.php
waldyrious: Ok, makes sense. If you'd like, I can make that change in a separate PR after this one, since indeed it is orthogonal to these changes.

waldyrious:
> I don't know what the default in the UI is (kinda hard to find a file with 25+ revs)

I was talking about history for regular pages, but that was just for a comparison point.

> Not retrieving more data than necessary in the initial API request, i.e. the default current revision instead of an arbitrarily larger number.

I agree with this, but it seems to me (and correct me if I'm wrong) that to get a particular (non-latest) revision, one would have to call `getHistory` before calign `getRevision(2)`. Is that the case? And if so, couldn't we make it so that, if `getRevision` receives a particular revison number parameter, it automatically fetches the corresponding data via a `getHistory` call?

To be honest, I'm not 100% sure I internalized how these methods are set up to interact with each other, so I may be saying something that is already the case -- let me know.

Xymph:
> to get a particular (non-latest) revision, one would have to call getHistory before caling getRevision(n). Is that the case?

Yes, hence that [usage example](https://github.com/hamstar/Wikimate/pull/76/files/d7cf8a4204192a3026de6db2b8a0074f99ae9555..d985310a4d80fe445daa23a031eecae8a199998d#diff-0f23af2f696313fe32c17a0c7281cb57R250).

> couldn't we make it so that, if `getRevision` receives a particular revison number parameter, it automatically fetches the corresponding data via a `getHistory` call?

Theoretically, yes, but revision number counted from where? The current (latest) revision or one within a selected timestamp range? How would you specify a limit and/or timestamp range from `getRevision` to `getHistory` if the former doesn't take those as input parameters? Would this still enable the user to access the MW API functionality in a straightforward manner?

To answer myself (but the questions weren't just rhetorical), I don't think so. I think this approach would mix up actions (fetch data from the API vs. returning cached data via getters) more than the rest of Wikimate does, and may make behavior of `getRevision` more confusing, perhaps less predictable. Currently the granular actions of `getHistory` followed by `getRevision` (or any other getter [if that's implemented](https://github.com/hamstar/Wikimate/pull/76#issuecomment-263122509)) indeed require the user to take more steps (well, one), but at least [s]he remains in control over every step and it's clear what outcome can be expected all the way.

That is actually entry 3 in my rationale which I reasoned through for myself a week ago when I wrote this code, but didn't remember right away this morning.

> To be honest, I'm not 100% sure I internalized how these methods are set up to interact with each other

I understand, that took me a while too, so please take further time for that as needed. I'm still not sure this history/revision approach is the only/best one to meet the aforementioned goals, but after having written the code I am probably not able to see another one. It would take someone else with thorough understanding of overall model and underlying code to point it out if it exists.

Wikimate.php
Xymph: Sure, please do.

waldyrious:
Considering that:
1. I trust your judgment, both overall given your past contributions, and in this particular change given that you've actually written the code to perform these actions; and
2. the advantage of having this feature at all vastly outweighs the mild inconvenience of a (possibly, not certainly) non-optimal interface

...I believe the clear right path forward is to merge these changes now, and let eventual improvements be integrated when/if the time comes, by those who happen to come up with a better approach.

I will thus merge the PR and open another one updating the example formats as mentioned above. Really nice work! 👍

Xymph:
Thanks. :smile: But what about my remaining open question? :wink:

waldyrious:
Sorry, which one do you mean, exactly? I'm not sure.

Xymph:
The one re-linked just to make sure... but not sure enough apparently. :stuck_out_tongue_winking_eye: 

> or any other getter [if that's implemented](https://github.com/hamstar/Wikimate/pull/76#issuecomment-263122509)

waldyrious:
Oh, right. I forgot about that one :P

Yes, I think that makes sense. In fact, the local complexity would increase, but the overall complexity would decrease since the usage would be more consistent.

Issue #79, opened 2017-01-05
WikiFile: add getAnon method


Xymph:
I discovered by chance that for a file uploaded by an anonymous user, an extra `anon` flag is returned with the imageinfo properties by the MW API. It indicates that the `user` field, containing the IP address, is not a username resembling an IP but an actual anon IP. The `getAnon` method should make this flag accessible to WikiFile users.

Xymph:
See #80.

PR #80, opened 2017-01-06
Add WikiFile::getAnon method, update changelog

Xymph:
See #79. Very straightforward, nothing more to say.


waldyrious:
Indeed 👍

PR #81, opened 2017-02-03
Prepare for v0.12.0 release

Xymph:
I've programmed against and used the current codebase for quite a while, encountered no problems, and thought of no further changes to make. So I figure it's time to wrap it up as a new stable release. Agreed, @waldyrious ?

PR #83, opened 2017-03-11
Create GOVERNANCE.md

waldyrious:
see https://github.com/hamstar/Wikimate/pull/81 for context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Since v0.10.0 this project adheres to [Semantic Versioning](http://semver.org/) 
 
 ### Unreleased version
 
+#### Added
+
+* New GOVERNANCE.md file to explicitly codify the project management principles and provide guidelines for maintenance tasks ([#83])
+
 #### Fixed
 
 * Prevented PHP notice in `WikiFile::getInfo()` for moved or deleted file. ([#85])
@@ -89,5 +93,6 @@ Since v0.10.0 this project adheres to [Semantic Versioning](http://semver.org/) 
 [#76]: https://github.com/hamstar/Wikimate/pull/76
 [#78]: https://github.com/hamstar/Wikimate/pull/78
 [#80]: https://github.com/hamstar/Wikimate/pull/80
+[#83]: https://github.com/hamstar/Wikimate/pull/83
 [#85]: https://github.com/hamstar/Wikimate/pull/85
 
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,0 +1,28 @@
+The Wikimate project strives for a transparent and welcoming maintenance process.
+Below are some guidelines for current and prospective maintainers.
+Anyone is welcome to propose changes to this document via issues or pull requests.
+
+Regular contributors are to be given collaborator status
+(which includes commit access, the ability to merge PRs, make releases, etc.).
+Note that at the moment this can only be done by the repository owner, @hamstar.
+Contributors are encouraged to ask for collaborator access once you have had a few accepted PRs, and would like to help maintaining the repository.
+
+## Guidelines for commits and pull requests
+
+1. **Every change** (except for very minor ones such as spelling fixes) **should be made via pull requests**,
+   rather than as direct commits to the repository.
+2. **Maintainers should not merge their own pull requests**.
+   This allows every change to be validated by at least another maintainer
+3. Pull requests should comprise a single feature or bugfix.
+   **Unrelated changes should be sent as separate PRs.**
+   The exception are minor code cleanup changes,
+   which can be included (as a separate commit) in the PR that prompted them.
+4. **Commits should be atomic** (as small as possible while still representing a self-consistent set of changes)
+   and have descriptive commit messages.
+5. **Every PR should include a CHANGELOG.md entry**.
+   This makes it much easier to prepare releases,
+   and allows the change author to properly summarize it.
+
+## Process for releasing a new version of Wikimate
+
+[TODO]
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -4,13 +4,14 @@ Anyone is welcome to propose changes to this document via issues or pull request
 
 Regular contributors are to be given collaborator status
 (which includes commit access, the ability to merge PRs, make releases, etc.).
-Note that at the moment this can only be done by the repository owner, @hamstar.
-Contributors are encouraged to ask for collaborator access once you have had a few accepted PRs, and would like to help maintaining the repository.
+Note that at the moment this can only be done by the repository owner, [@hamstar](https://github.com/hamstar).
+Contributors that have had a few PRs merged are encouraged to ask for collaborator access,
+if they would like to help maintaining the repository.
 
 ## Guidelines for commits and pull requests
 
 1. **Every change** (except for very minor ones such as spelling fixes) **should be made via pull requests**,
-   rather than as direct commits to the repository.
+   rather than as direct commits to the `master` branch.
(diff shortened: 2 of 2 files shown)


waldyrious:
Do not merge yet. I will add content to the release section shortly.

GOVERNANCE.md
```diff
@@ -25,4 +26,25 @@ Contributors are encouraged to ask for collaborator access once you have had a f
 
 ## Process for releasing a new version of Wikimate
 
-[TODO]
+A PR should be created with all relevant changes to update the repository for the upcoming release.
+(See [#81](https://github.com/hamstar/Wikimate/pull/81) for an example.)
+It should apply the following actions:
+
+1. Change the "Unreleased" heading in the CHANGELOG.md file
```
Xymph: The phrase we previously agreed on and used was "Current development".

GOVERNANCE.md
```diff
@@ -0,0 +1,28 @@
...
+(which includes commit access, the ability to merge PRs, make releases, etc.).
+Note that at the moment this can only be done by the repository owner, @hamstar.
+Contributors are encouraged to ask for collaborator access once you have had a few accepted PRs, and would like to help maintaining the repository.
+
+## Guidelines for commits and pull requests
+
+1. **Every change** (except for very minor ones such as spelling fixes) **should be made via pull requests**,
+   rather than as direct commits to the repository.
+2. **Maintainers should not merge their own pull requests**.
+   This allows every change to be validated by at least another maintainer
```
Xymph: +period

Xymph:
Two minor textual fixes, otherwise this looks fine to me. :+1:

GOVERNANCE.md
waldyrious: I can't recall how detailed the previous discussion was (I did a search but couldn't find it), but the rationale here is that if we're claiming/recommending [Keep a Changelog](http://keepachangelog.com/) compliance, then we should use the heading they propose, i.e. "Unreleased" (which IMO is also more explicit than "Current development", and fits better with the other headings), unless there are strong arguments to deviate from it.

Xymph: [Discussion](https://github.com/hamstar/Wikimate/pull/45#issuecomment-252443635) led to issue #48, and the resulting commits were approved without further discussion it seems. So I guess our agreement was more implicit than I remembered.

Does adhering to a standard mean that we have no creative freedom to deviate in how we word something? [Precedent](https://gcc.gnu.org/wiki/GFortran/News#gfortran_7_.28current_development_version.29) [does](http://docs.nagvis.org/1.3/en_US/changelog.html) [exist](http://pyhrf.org/changelog.html), although I've not found many more examples.
Anyway, I like "current development" better because it has a more positive and active connotation. Many words starting with "un" feel less so.

How exactly do you mean "fits better with the other headings"?

waldyrious: > [Discussion](https://github.com/hamstar/Wikimate/pull/45#issuecomment-252443635) led to issue #48, and the resulting commits were approved without further discussion it seems. So I guess our agreement was more implicit than I remembered.

Ah, yes, thanks for the links :)

> Does adhering to a standard mean that we have no creative freedom to deviate in how we word something? [Precedent](https://gcc.gnu.org/wiki/GFortran/News#gfortran_7_.28current_development_version.29) [does](http://docs.nagvis.org/1.3/en_US/changelog.html) [exist](http://pyhrf.org/changelog.html), although I've not found many more examples.

I'd say we can deviate from the standard, if sufficient motive exists.

> Anyway, I like "current development" better because it has a more positive and active connotation. Many words starting with "un" feel less so.

I can agree with this argument, but I'd rather we picked a different section name that still conforms to it. I'm afraid "Current development" may not be sufficiently clear. It certainly feels slightly awkward to my (non-native English speaker) ears, for the meaning it is intended to have.

> How exactly do you mean "fits better with the other headings"?

I meant that all headings are version this, version that; so it would logically make sense (IMO) that the top heading is "upcoming version", "next release", "unreleased [version]", or something to that effect. Do any of these (or a variation thereof) appeal to you?

Xymph: I see, thanks for the clarification. TBH, none appeal much to me, nor appall me, but it took me a while to realize (or remember from the time when I choose "Current development") why.

On Git(Hub) the formal meaning of "release" is to tag the current state of affairs into something fixed, no longer in flux. But to me the very act of putting something online is already "releasing" it, in the practical meaning. That's how I worked with my previous, non-gitted projects; I didn't put anything online until it had enough quality and quantity to be version-able.

So "unreleased" or (worse) "next release" feels like a misnomer because the content under that heading is already _in_ the master branch, is already "out there". I guess I'm still not quite used to GitHub's meaning of "release". To my equally non-native ears my original choice of header sounds like a adequate description of the entries below it, which can remain in a state of flux until tagged (see the number of PRs until WikiFile was ready).

But if you're still feeling awkward about it, then I'm fine with simply sticking to the Keepachangelog standard verbatim. We are, after all, working on GitHub now. :smile:

waldyrious: I see what you mean, but yeah, I guess people do end up getting used to the github model. I never even considered any other interpretation until you brought this up :)

Now that I've thought about it more explicitly, I do believe this is a better way of working with FOSS, as it promotes openness and transparency of the development process: unpolished changes go in separate branches (and WIP commits can be easily squashed, amended, etc); changes ready for review by maintainers go to the PR stage; changes considered ready are merged into master; and changes that have been on master for a while (hopefully tested by the most intrepid users who prefer not to wait for releases) are released with a stable (immutable) tag, downloadable as an archive, bundled with release notes, supported by the maintainers, etc.

Given this, I don't think we should word the top section based on the premise than master is a "release". If none of the options appeal to you, I agree, let's stick to what Keep a Changelog suggests.

GOVERNANCE.md
waldyrious: Fixed. Nice attention to detail, as always ;)

GOVERNANCE.md
Xymph: Alright, the "version x.y.x" header pattern makes "Unreleased version" a little more appealing than the others. If you agree, I'll update #85.

waldyrious: Agreed 👍

waldyrious:
@Xymph if everything looks good, please merge (squashing sounds sensible here, I suppose) at your leisure.

Xymph:
I think a Changelog entry for the new file (under an "Added" heading) would make sense. To avoid a conflict, best review/merge #85 first.

waldyrious:
Just to confirm, by the new file do you mean GOVERNANCE.md?

Xymph:
Uhm, yes, what other new file was added in this "Create" PR? :stuck_out_tongue_closed_eyes:

waldyrious:
@Xymph rebased and updated CHANGELOG.md -- thanks for the prompt.

CHANGELOG.md
```diff
@@ -89,5 +93,6 @@ Since v0.10.0 this project adheres to [Semantic Versioning](http://semver.org/)
 [#76]: https://github.com/hamstar/Wikimate/pull/76
 [#78]: https://github.com/hamstar/Wikimate/pull/78
 [#80]: https://github.com/hamstar/Wikimate/pull/80
+[#85]: https://github.com/hamstar/Wikimate/pull/83
```
Xymph: -> [#83]

waldyrious: Argh. Good catch.

Xymph:
Alright, your turn to wrap one up. :+1:

waldyrious:
Took me quite a while, but I can finally check this off my todo list! Thanks for the review.

PR #85, opened 2017-03-12
Fix PHP notice in WikiFile::getInfo() (fixes #84)

### LINKED ISSUE #84: WikiFile::getInfo on deleted file gives PHP Notice
I discovered by chance, when a file existed but was subsequently deleted or moved to another path, that the MW API does not return the `missing` property for the requested path anymore. I guess it only does that when the path never existed in the first place.

In the following `getInfo()` excerpt:
```
        if (!isset($page['missing'])) {
                // Update the existence if the file is there
                $this->exists = true;
                // Put the content into info & history
                $this->info    = $page['imageinfo'][0];
                $this->history = $page['imageinfo'];
        }
```
this results in PHP notices:
```
PHP Notice:  Undefined index: imageinfo in /usr/local/share/wikimate/Wikimate.php on line 1104
PHP Notice:  Undefined index: imageinfo in /usr/local/share/wikimate/Wikimate.php on line 1105
```
That's logical as the `imageinfo` property array is not present for a non-existing file.

Also, a changelog typo regrettably slipped through the v0.12.0 release: it was [named 0.11.0](https://github.com/hamstar/Wikimate/commit/2f7f77687435f4d2469c362946718637af5adeca#diff-4ac32a78649ca5bdd8e0ba38b7006a1e). Even two people can be oblivious to the obvious. :astonished:

Xymph:
See #84.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -1097,7 +1097,8 @@ class WikiFile
 
 			$this->edittoken = $page['edittoken'];
 
-			if (!isset($page['missing'])) {
+			// Check that file is present and has info
+			if (!isset($page['missing']) && isset($page['imageinfo'])) {
 				// Update the existence if the file is there
 				$this->exists = true;
 				// Put the content into info & history
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,11 @@
 
 Since v0.10.0 this project adheres to [Semantic Versioning](http://semver.org/) and [Keep a Changelog](http://keepachangelog.com/).
 
-### Current development
+### Unreleased version
 
-* Fix PHP notice in `WikiFile::getInfo()` for moved or deleted file. ([#85])
+#### Fixed
+
+* Prevented PHP notice in `WikiFile::getInfo()` for moved or deleted file. ([#85])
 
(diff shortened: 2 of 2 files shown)


Wikimate.php
```diff
@@ -1097,7 +1097,7 @@ public function getInfo($refresh = false, $history = null)
 
 			$this->edittoken = $page['edittoken'];
 
-			if (!isset($page['missing'])) {
+			if (!isset($page['missing']) && isset($page['imageinfo'])) {
```
waldyrious: Do you think a comment would be warranted above this line? I'm not sure the behavior you described in #84 is intuitive: 

> when a file existed but was subsequently deleted or moved to another path, that the MW API does not return the `missing` property for the requested path anymore. I guess it only does that when the path never existed in the first place.

Xymph: Yes, good point - done.

waldyrious:
LGTM. Feel free to do any commit cleanup if you feel it's worth it, and merge whenever you're ready.

Xymph:
I fixed up the comment with the code change, but preserved the reference to the changelog discussion. That's a wrap. :)

PR #97, opened 2021-03-27
Small intro correction in README

Xymph:
Update class count in intro.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 Wikimate is a wrapper for the MediaWiki API that aims to be very easy to use.
-It currently consists of two classes:
+It currently consists of three classes:
 
 * **Wikimate** – Serves as a loader and manager for different wiki objects (e.g. pages).
 * **WikiPage** – Provides an interface to getting/editing pages or sections of them.


waldyrious:
This was squash-merged, right? AFAIK GitHub doesn't do fast-forward merges, so a new commit is always created:

```
Before merge:                  After merge:

                               D (master)
  C (Xymph:master)             | C (Xymph:master)
 /                             |/
B (master)          --->       B
|                              |
A                              A
```

That is, commit `D` is always created even if its content is exactly the same as `C`.

Furthermore, when you pulled the latest changes from the original repo into your fork, GitHub created a merge commit, since the graphs had "diverged" (as far as git can tell) and your fork's commit graph couldn't be fast-forwarded to match the upstreams's graph. So this is what happened:

```
Before pull:                   After pull:

                               E (Xymph:master)  <-- merge commit
                               |\
D (master)                     D | 
| C (Xymph:master)             | C 
|/                   --->      |/
B                              B
|                              | 
A                              A
```

I think the best way to avoid this is to submit PRs from dedicated branches in your fork, rather than from `master`. I suspect that pulling the latest changes from upstream into your fork would then not create merge commits, but simply fast-forward your graph to include the new upstream commits.

For now, the best way to resolve things is to force-set the master branch in your fork to point to the same commit as upstream master:

    git branch -f master 3c8833a

That should get the graphs looking like this:

```
D (master) (Xymph:master)
|
B
|
A
```

Hope this helps! Let me know if there's anything I should clarify.

Xymph:
TBH, I don't recall how I committed/merged the README change in March, but I definitely didn't create a branch for it in my local repo (didn't seem worth the trouble for a one-word change). Advice noted about doing that in the future.

And {sheepish grin} I guess I just made it worse by hastily merging and pulling #98 before resolving this here problem. The git command is denied with "Cannot force update the current branch".

I wouldn't mind deleting my entire fork and starting over, if that is the easiest way out by now.

waldyrious:
> The git command is denied with "Cannot force update the current branch".

I don't think that's related to the merge of #98. I believe git might not allow you to force-update the current branch. If I'm correct, this should work:

```console
$ git checkout HEAD            # "Step out" of the `master` branch
$ git branch -f master c48284c # Note: the ref is no longer 3c8833a since the merge of #98.
$ git checkout master          # Step back into `master`
```

Xymph:
The first checkout warns: "Your branch is ahead of origin/master by 4 commits."
The branch command gives the same fatal error as before.
The second checkout says I'm already on 'master'.

Does that help?

waldyrious:
It does, sorry. I misremembered that `git checkout HEAD` does not actually get you into detached head state. Try `git checkout HEAD~` instead as the first command. That will put you on a commit other than the tip of the `master` branch, and then you should be able to force-update it since you're no longer "sitting" on it.

Xymph:
Thanks, now it worked, and master and my graphs are clean. You're a life saver! :)
I'll try not to get myself in this kind of trouble again.
(not making promises about other kinds...)

waldyrious:
I'm glad it worked! It was good for me as well to brush up on some concepts :)

By the way, I always recommend [this talk](https://www.youtube.com/watch?v=1ffBJ4sVUb4) ("Git For Ages 4 And Up") which helped me build intuition for git's graph. Once the model is clear in your head, it becomes possible to translate obscure-looking commands into "physical", visual operations in a graph. `git log --graph` is also crucial to understand the state of the repository.

Xymph:
Yes, I viewed that when you first recommended it (and parts of it again later), but after all this time it has faded into memory a bit, and understanding what was wrong with my repo wasn't so trivial anymore. Thanks again for your time.

Issue #92, opened 2019-01-17
Add support for passing maxlag to the API


addshore:
Someone is currently using this code to make requests to the www.wikidata.org api.
They are not currently passing a maxlag value to the api (which they should).
I had a quick skim and it looks like support needs to be added.

Xymph:
Thanks for the heads-up. Documentation: https://www.mediawiki.org/wiki/Manual:Maxlag_parameter

If someone submits a PR to implement this, I may be able to review it soon, but I won't have time to develop it myself for quite a while.

Xymph:
This is now implemented in PR #112. @addshore, if you're able to shed some reviewing light on it, that would be welcome.

Xymph:
@addshore This is released in v0.14.0, perhaps you can note it on Phabricator.

PR #100, opened 2021-06-19
Modernize token handling (fixes #99)

### LINKED ISSUE #99: Deprecated CSRF token parameters
- https://lists.wikimedia.org/hyperkitty/list/wikitech-l@lists.wikimedia.org/thread/IMP43BNCI32C524O5YCUWMQYP4WVBQ2B/
- https://github.com/hamstar/Wikimate/blob/3c8833abd5cfed36beb0e04d024727bdfc7baea9/Wikimate.php#L537

Xymph:
The action token handling is [deprecated](https://www.mediawiki.org/wiki/API:Tokens_(action)) since MW v1.24. This PR restructures all three Wikimate classes to use the current [meta tokens approach](https://www.mediawiki.org/wiki/API:Tokens). That includes the login token, which was only introduced in MW v1.27. Wikimate deployments prior to that version can continue to use the v0.12.0 tag.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -103,7 +103,7 @@ class Wikimate
 		if ($this->debugMode) {
 			echo "Token request:\n";
 			print_r($details);
-			echo "Token request response:\n";
+			echo "Token response:\n";
 			print_r($tokenResult);
 		}
 
@@ -126,10 +126,16 @@ class Wikimate
 	{
 		//Logger::log("Logging in");
 
+		// Obtain login token first
+		if (($logintoken = $this->token(self::TOKEN_LOGIN)) === false) {
+			return false;
+		}
+
 		$details = array(
 			'action' => 'login',
 			'lgname' => $username,
 			'lgpassword' => $password,
+			'lgtoken' => $logintoken,
 			'format' => 'json'
 		);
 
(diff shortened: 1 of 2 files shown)


Xymph (commented):
Re. cc9bc21, these Logger lines were introduced way back in v0.4 but were never active as far as I could find, and it's not even clear which Logger library they relate too. Hence this little house-keeping.

Xymph (commented):
This large(r) commit updates all three classes in order not to leave Wikimate in an inconsistent state between commits.

Xymph (commented):
This method, in Wikimate tradition, is kept simple and supports only the two token types needed elsewhere in the library. So no patrol, rollback, watch, etc. Nor the option to request more than one token at once ("watch|patrol"), we can cross that bridge if and when we get to it.

Wikimate.php
```diff
@@ -852,16 +866,20 @@ public function setText($text, $section = null, $minor = false, $summary = null)
...
-				'intoken' => 'edit',
+				'curtimestamp' => 1,
 			);
 
 			$r = $this->wikimate->query($data);
 
-			$page = array_pop($r['query']['pages']); // Get the page
-
-			$this->starttimestamp = $page['starttimestamp']; // Update the starttimestamp
+			// Check for errors
```
Xymph, line 874: This adds error checking to the preceding query(), analog to such invocations elsewhere.

Xymph:
@waldyrious As before (well, in 2016-2017 :) ) a second pair of eyes on these commits would be appreciated.

waldyrious:
> Wikimate deployments prior to that version can continue to use the v0.12.0 tag.

Don't you mean "MediaWiki deployments..."?

Wikimate.php
waldyrious: Any reason why this comment was not added to the code itself?

waldyrious:
Awesome job in separating the commits! I can't review the code to any meaningful depth, since I haven't used PHP nor the MW API in quite a while. The superficial aspects (commit messages and separation, code formatting, wording in comments, etc.) all look good to me. I'd only repeat one of the comments I've left inline, regarding this PR-level comment:

> Re. 5044fd7, this method, in Wikimate tradition, is kept simple and supports only the two token types needed elsewhere in the library. So no patrol, rollback, watch, etc. Nor the option to request more than one token at once ("watch|patrol"), we can cross that bridge if and when we get to it.

Here too, wouldn't it make sense to add the comment to the code itself? It would be cumbersome for someone reading the code in the future to have to locate a pull request discussion thread to obtain context for why that method is implemented that way.

Xymph:
>> Wikimate deployments prior to that version can continue to use the v0.12.0 tag.
>
>Don't you mean "MediaWiki deployments..."?

Sorry that wasn't clear, I meant: Wikimate usage against MW deployments prior...

Wikimate.php
Xymph: My remark should have been a (multiline) comment for new code lines 874 through 880. All other places where query() was called were followed by the code now also in the new lines. Not sure why it was previously missing, this was old code, but I don't think it's meaningful to spend more in-code comments on it.

Xymph:
> Here too, wouldn't it make sense to add the comment to the code itself? It would be cumbersome for someone reading the code in the future to have to locate a pull request discussion thread to obtain context for why that method is implemented that way.

Good point, how about these two sentences?
```
   * For now this method, in Wikimate tradition, is kept simple and supports
   * only the two token types needed elsewhere in the library.  It also
   * doesn't support the option to request multiple tokens at once.
```
Thanks for the other remarks, the feedback is welcome.

If this is okay and there are no other items, then I'll wrap it up and merge-commit (to preserve individual commit messages).

Xymph:
Unless you are very busy, @waldyrious, perhaps you missed my follow-ups because I didn't include a ping, like now.

Xymph:
I guess it's good enough, so wrapping up now.

Wikimate.php
waldyrious: Got it, thanks for the context.

waldyrious:
Hey @Xymph — I was actually out for a few days, sorry I wasn't able to respond in time. In any case, yeah, I didn't have any additional feedback and the changes you've made look good to me<sup>*</sup>, so here's my late LGTM! 😛

<sub>* I personally would have tended to include a bit more of the context you had in the PR comment ("So no patrol, rollback, watch, etc." and the example for multiple tokens, i.e.  ("watch|patrol")) into the code comment, but as you say, we need to strike a balance between completeness and verbosity.</sub>

Xymph:
That's okay, I saw some activity in your heatmap early this week so figured you were busy elsewhere, but thanks for the LGTMs here and on #101. :)

For the token() method, I can add the link to the [API page](https://www.mediawiki.org/wiki/Special:MyLanguage/API:Tokens) so readers can see examples there of what isn't supported?

waldyrious:
> I saw some activity in your heatmap early this week

That's my plight with GH notifications 😅 sometimes I am able to come by just to report a quick issue or respond to a comment that requires no further action, but I always fear that other people who are waiting on me (for things that do require a more attentive look) will think I'm ignoring them.

Furthermore, I do get mobile notifications when I'm mentioned, but if I do open them to respond, the thread leaves my notifications inbox. So it's not even easy to drop a quick "I'm aware of this but can't take care of it right away" message without messing with my ability to keep track of the threads I need to come back to later...

> For the token() method, I can add the link to the [API page](https://www.mediawiki.org/wiki/Special:MyLanguage/API:Tokens) so readers can see examples there of what isn't supported?

That's a good idea, yeah :)

Xymph:
> > For the token() method, I can add the link to the [API page](https://www.mediawiki.org/wiki/Special:MyLanguage/API:Tokens) so readers can see examples there of what isn't supported?
> 
> That's a good idea, yeah :)

Ok, done in fafa2d8

reedy:
Thanks! I was coming here to ask about doing this for https://phabricator.wikimedia.org/T280806, but now there's a release... Just need to wait for the various bots to update :)

Xymph:
> Thanks! I was coming here to ask about doing this for https://phabricator.wikimedia.org/T280806, but now there's a release... Just need to wait for the various bots to update :)

Our pleasure. The next release will address another Phabricator ticket per issue #92 / PR #112. Btw, we're looking for additional reviewing eyes there, in case that's something you could spend a little time on (but no problem if you can't).

reedy:
Just need to get people to stop using versions 0.11 and 0.12 now ;)

PR #101, opened 2021-06-28
Added debug logging of MediaWiki requests and responses

Xymph:
During my own development and testing of Wikimate and my own client scripts, I found these request/response debug logs rather useful, so adding them to the repo.

I can imagine that this is too much logging in some situations and that we'd like debug levels, but that would require further changes and possibly an API-breaking change as $debugMode is currently a boolean.

Would anyone have additional thoughts on this? @waldyrious perhaps?

Also applying some comment improvements that aren't worth a separate branch.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -34,7 +34,7 @@ class Wikimate
 	protected $debugMode = false;
 
 	/**
-	 * Create a new Wikimate object.
+	 * Creates a new Wikimate object.
 	 *
 	 * @return  Wikimate
 	 */
@@ -49,7 +49,7 @@ class Wikimate
 	}
 
 	/**
-	 * Set up a Requests_Session with appropriate user agent.
+	 * Sets up a Requests_Session with appropriate user agent.
 	 *
 	 * @return  void
 	 */
@@ -215,7 +215,7 @@ class Wikimate
 	}
 
 	/**
-	 * Get or print the Requests configuration.
+	 * Gets or prints the Requests configuration.
 	 *
 	 * @param   boolean  $echo  Whether to echo the options
 	 * @return  array           Options if $echo is false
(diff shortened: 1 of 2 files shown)


waldyrious:
Sorry I couldn't review this on time. In any case, I did take a look now, and everything looks good on my end. Nice commit separation, as always :)

Wikimate.php
```diff
@@ -283,8 +291,16 @@ public function parse($array)
...
 
+		if ($this->debugMode) {
+			echo "parse GET parameters:\n";
+			echo http_build_query($array) . "\n";
+		}
 		$apiResult = $this->session->get($this->api.'?'.http_build_query($array));
 
+		if ($this->debugMode) {
+			echo "parse GET response:\n";
+			print_r(json_decode($apiResult->body));
```
Xymph, line 302: This is trying to JSON-decode PHP format. Fixed in a1a44fa

Xymph:
I've made good progress on #113 and in retesting some client scripts with debug logging, I've grown to feel there's no real benefit to debug levels anymore. My original thinking there was related to token requests vs. query/other requests, because each edit/delete/upload refetched the CSRF token. Level 1 (e.g.) would then keep token requests out of the logging, and level 2 include them. But that no longer happens since #115.

Now with centralized logging of GET requests to and GET/POST responses from the API, I find to analyze a problem I simply need to check all requests in a flow. And analyzable problems are more often in the Wikimate classes than in client scripts. So unless and until other users express more refined needs for debug logging, I think it's better to keep this Wikimate-simple. :smiley: 

Btw, the info/warning/error grouping doesn't apply to this library: it doesn't output any informational messages, and errors/warnings (from the API or Wikimate itself) are returned via `getError()` or exceptions.

waldyrious:
Sounds good to me. Simple is good :)

PR #102, opened 2021-07-01
Prepare for v0.13.0 release

Xymph:
I'm working on maxlag support which brings about significant code changes, so it's good to first create a stable release of where we are now.

If @waldyrious continues to be unavailable that's okay, but in absence of another active collaborator that means I'll have to keep merging my own PRs, in spite of this repo's governance guidelines. I intend to wrap this up on Sunday July 4.

PR #105, opened 2021-07-04
Update governance guidelines

waldyrious:
- Add missing content to CHANGELOG.md (version 0.9, and reference links for #100 and #101)
- Add a provision to unblock PRs when only one maintainer is available
- Change title of CHANGELOG section for untagged changes, and define default content for it
- Add dates for released versions in CHANGELOG.md

See the relevant discussions in #102 for context.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -35,7 +35,7 @@ A PR should be created with all relevant changes to update the repository for th
 It should apply the following actions:
 
 1. Change the "Upcoming version" heading in the CHANGELOG.md file
-   to the appropriate version name (e.g. "Version 1.2.3")
+   to the appropriate version name and date (e.g. "Version 1.2.3 - 2020-12-31")
    and add a new "Upcoming version" section heading above it,
    with the contents "No changes yet.";
 2. Edit the README and replace all references
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -24,7 +24,7 @@ if they would like to help maintaining the repository.
 4. **Commits should be atomic**
    (as small as possible while still representing a self-consistent set of changes)
    and have descriptive commit messages.
-5. **Every PR should include a CHANGELOG.md entry**.
+5. **Every PR should include a `CHANGELOG.md` entry**.
    This makes it much easier to prepare releases,
    and allows the change author to properly summarize it.
 
@@ -34,14 +34,14 @@ A PR should be created with all relevant changes to update the repository for th
 (See [#81](https://github.com/hamstar/Wikimate/pull/81) for an example.)
 It should apply the following actions:
 
-1. Change the "Upcoming version" heading in the CHANGELOG.md file
+1. Change the "Upcoming version" heading in the `CHANGELOG.md` file
    to the appropriate version name and date (e.g. "Version 1.2.3 - 2020-12-31")
    and add a new "Upcoming version" section heading above it,
    with the contents "No changes yet.";
 2. Edit the README and replace all references
    to the previous version number and release date
    with the corresponding data for the new version;
-3. Update all version references in Wikimate.php
+3. Update all version references in `Wikimate.php`
    to the new version.
 
 Once this PR is merged, a new release should be created
@@ -52,4 +52,4 @@ The version tag and the title of the release notes should be in the format v1.2.
 (following [SemVer](http://semver.org/) conventions
 to determine which part of the version number to increase).
 The body of the release notes should be a summary of the contents
-of the relevant section in CHANGELOG.md.
+of the relevant section in `CHANGELOG.md`.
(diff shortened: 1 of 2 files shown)


GOVERNANCE.md
```diff
@@ -33,12 +35,13 @@ A PR should be created with all relevant changes to update the repository for th
 It should apply the following actions:
 
 1. Change the "Unreleased version" heading in the CHANGELOG.md file
```
waldyrious, line 37: @Xymph any thoughts about renaming this to "Upcoming version" or "Next version", as I suggested [here](https://github.com/hamstar/Wikimate/pull/102#discussion_r662137004)?

GOVERNANCE.md
```diff
@@ -33,12 +35,13 @@ A PR should be created with all relevant changes to update the repository for th
...
-   to the appropriate version name (e.g. "Version 1.2.3")
-   and add a new "Unreleased version" section heading above it;
+   to the appropriate version name and date (e.g. "Version 1.2.3 - 2020-12-31")
```
waldyrious, line 38: I believe it would make sense to replicate the changes from e2c026c42639c998d399513126bf918d7aaea153 in this PR, so that they're introduced in lockstep with these guideline changes. WDYT @Xymph?

GOVERNANCE.md
Xymph: "Upcoming version" works for me as well.

GOVERNANCE.md
Xymph: I'm not really biased either way, but if you are splitting off commits from #102 anyway, I'm fine with that commit going into this PR too. Along with the initial header change per my comment above.

Xymph:
No further comments on the committed changes, they look fine to me.

GOVERNANCE.md
waldyrious: I can't cleanly cherry-pick that commit into this branch, because it includes adding the section header for the new release (which should remain in #102), but I'll recreate the parts that need to be moved and will mark you as co-author of the commit :)

Xymph: Fine, if I need to make changes to my branch locally, and pull/push stuff around, please detail any steps.

GOVERNANCE.md
waldyrious: Done in 1a9ae641cced375474d378f10a0654d562575623 :)

GOVERNANCE.md
waldyrious: Yes, no worries. In fact, I should be able to update your branch myself. Then, if it looks good to you on GitHub, you will be able to pull the updated branch to your local clone as well.

waldyrious:
OK, changes made. Please check the updated commits.

Xymph:
Almost everything looks fine to me, the only previous change I don't see anymore is the formatting of `Wikimate.php` in bullet 3 of the last section.

Btw, remember to add #105 to the Changelog for v0.13.0.

waldyrious:
The formatting reversal was intentional — the other filenames in that document are not code-formatted, so I reverted it for consistency. I can change them all to be code-formatted, though. WDYT?

As for adding a changelog entry — OMG, of course! I'll add a new commit for that.

Xymph (approved):
Looking good now.

Xymph (commented):
Re. f3d4338: The Governance doc is a new addition in v0.13.0, so instead of a separate entry under Changes, wouldn't it be logical to simply add this PR to that first entry, like "(#83, #105)"? Similar to older changelog entries combining multiple PRs.

waldyrious:
Good point. I forgot we could combine multiple PRs into the same entry. Will change.

Xymph:
Great, looks ready to wrap up.

waldyrious:
Great :) Please go ahead at your convenience.

GOVERNANCE.md
```diff
@@ -34,9 +34,10 @@ A PR should be created with all relevant changes to update the repository for th
 (See [#81](https://github.com/hamstar/Wikimate/pull/81) for an example.)
 It should apply the following actions:
 
-1. Change the "Unreleased version" heading in the CHANGELOG.md file
+1. Change the "Upcoming version" heading in the CHANGELOG.md file
    to the appropriate version name (e.g. "Version 1.2.3")
-   and add a new "Unreleased version" section heading above it;
+   and add a new "Upcoming version" section heading above it,
+   with the contents "No changes yet.";
```
Xymph, line 40: A belated and very tiny nitpick, but changelog entries normally don't end with a period, so this dummy entry shouldn't either.

waldyrious: I can see where you're coming from, but on the other hand the changelog entries are formatted as bullet list items (which commonly don't use punctuation) while this is a regular sentence. I don't think it would make sense to format it as a list item, btw.

If you agree, I'll go ahead and add the period in #102.

Xymph: Ah, indeed - agreed on both, uhm, points. :smile:

waldyrious: Great, let's resolve this thread then! I just pushed an update to #102.

PR #107, opened 2021-07-06
Support custom user agent for API requests (fixes #103)

### LINKED ISSUE #103: Allow custom user agent
It would be good to be able to set a custom user agent, so that clients using this library can abide by the [User-Agent policy](https://meta.wikimedia.org/wiki/User-Agent_policy).

I suggest adding a new method: `Wikimate::setUserAgent( string $userAgent ): void`

Xymph:
Adds get and set methods for the user agent string, with usage examples.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -189,6 +189,34 @@ class Wikimate
 	}
 
 	/**
+	 * Gets the user agent for API requests.
+	 *
+	 * @return  string  The default user agent, or the current one defined
+	 *                  by {@link Wikimate::setUserAgent()}
+	 */
+	public function getUserAgent()
+	{
+		return $this->useragent;
+	}
+
+	/**
+	 * Sets the user agent for API requests.
+	 *
+	 * In order to use a custom user agent for all requests in the session,
+	 * call this method before invoking {@link Wikimate::login()}.
+	 *
+	 * @param   string   $ua  The new user agent
+	 * @return  Wikimate      This object
+	 */
+	public function setUserAgent($ua)
+	{
+		$this->useragent = (string)$ua;
+		// Update the session
+		$this->session->useragent = $this->useragent;
+		return $this;
+	}
+
+	/**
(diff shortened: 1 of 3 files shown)


Wikimate.php
```diff
@@ -188,6 +188,34 @@ public function login($username, $password, $domain = null)
 		return true;
 	}
 
+	/**
+	 * Gets the user agent for API requests.
+	 *
+	 * @return  string  The default user agent, or the current one defined
+	 *                  by {@link Wikimate::setUserAgent()}
```
Xymph, line 195: A note ahead of the planned PHPDoc cleanup PR: this link isn't hyperlinked by my (possibly too old) phpdoc command. The login() link in the setUserAgent comments is, but no formatting variation attempts here will fix it in this comment. Letting this be for now.

Wikimate.php
```diff
@@ -188,6 +188,34 @@ public function login($username, $password, $domain = null)
...
+	 */
+	public function getUserAgent()
+	{
+		return $this->useragent;
+	}
+
+	/**
+	 * Sets the user agent for API requests.
+	 *
+	 * In order to use a custom user agent with all requests in the session,
```
waldyrious, line 205: Just a minor nit: I think this would read a bit more clearly like this:

```suggestion
	 * In order to use a custom user agent for all requests in the session,
```

or:

```suggestion
	 * In order to use a custom user agent in all requests of the session,
```

WDYT?

Xymph: Sure, I picked the former, see f6983f7.

Btw, this comment started as a note-to-self :wink:, as in a test script against a [wiki ](https://wiki.xaseco.org/) on my personal server (so I can access the Apache logs), I found that using setUserAgent too late caused the initial token/login requests to still have the default agent string. Figured it might as well be a reminder to other library users too.

waldyrious (approved):
LGTM!

Xymph:
Thanks, then please merge.

PR #108, opened 2021-07-06
Group sections and add table of contents in USAGE.md

Xymph:
Commits speak for themselves, I guess.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -349,6 +349,9 @@ $wiki->setUserAgent('Custom User Agent');
 $wiki->setUserAgent('Custom Prefix - ' . $useragent);
 ```
 
+In order to use a custom user agent for all requests in the session,
+call this method before invoking `Wikimate::login()`.
+
 #### Handling errors
 
 Did something go wrong?  Check the error array:
--- a/USAGE.md
+++ b/USAGE.md
@@ -1,5 +1,22 @@
 ## Usage
 
+- [Introduction](#introduction)
+- [Getting a page object](#getting-a-page-object)
+  * [Reading...](#reading)
+  * [Writing...](#writing)
+  * [Deleting...](#deleting)
+- [Getting a file object](#getting-a-file-object)
+  * [Downloading...](#downloading)
+  * [Uploading...](#uploading)
+  * [Accessing revisions...](#accessing-revisions)
+  * [Deleting...](#deleting-1)
+- [Other stuff](#other-stuff)
+  * [Running custom queries](#running-custom-queries)
+  * [Customizing the user agent](#customizing-the-user-agent)
+  * [Handling errors](#handling-errors)
+
+### Introduction
+
 In your script file (e.g. `index.php`), include the `globals.php` file,
 and create a new `Wikimate` object with the target wiki's API address.
 Then provide a username and password to Wikimate's `login` method,
@@ -52,7 +69,7 @@ echo $page->getNumSections();
 echo $page->getSectionOffsets();
 ```
 
-### Reading...
+#### Reading...
 
 You can get the text of the page by using the `getText()` method
 which returns the text that was obtained when the page object was created.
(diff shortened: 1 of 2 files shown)


Xymph:
One from the afterthought department: that setUserAgent comment would be a more [effective reminder](https://github.com/hamstar/Wikimate/pull/107#discussion_r664447951) if it's also in the Usage section?

USAGE.md
```diff
@@ -228,7 +245,7 @@ if ($file->exists()) die('Example image already exists');
 $result = uploadFromUrl('https://upload.wikimedia.org/wikipedia/en/a/a9/Example.jpg', 'Adopt Wiki example image');
 ```
 
-### History...
+#### History...
```
waldyrious, line 248: How about "Viewing the history", "Browsing history", or something along those lines?

And maybe the same for the "Other stuff" subsections: "Running custom queries", "Customizing the user agent", "Handling errors".

What do you think?

waldyrious:
> Commits speak for themselves, I guess.

IMO the message of the second commit should make a reference to the USAGE doc, just like the first one does. I know that we're grouping commit series together with merge commits, but it's best to reduce implicit content wherever possible.

> One from the afterthought department: that [setUserAgent comment](https://github.com/hamstar/Wikimate/commit/f6983f739f17a9b05de1dbb695578f2321038484#diff-915f1e0f55c2400eb73093b0452dbac786cbcb0a04a0f5b0db0e4adaf5b81fbbR205) would be a more [effective reminder](https://github.com/hamstar/Wikimate/pull/107#discussion_r664447951) if it's also in the Usage section?

Definitely. And this PR seems like a good place to add that change :)

waldyrious:
By the way, let's make sure to always use the [imperative mood](https://chris.beams.io/posts/git-commit/#imperative) in commit messages and PR titles. That's already the general practice, but a few exceptions (like those in this PR) have slipped by.

USAGE.md
Xymph: Good suggestions, adopted most of them.

Xymph:
> IMO the message of the second commit should make a reference to the USAGE doc, just like the first one does. I know that we're grouping commit series together with merge commits, but it's best to reduce implicit content wherever possible.

Hmm, no original commit message did, but I can add it to both?

> > One from the afterthought department: that [setUserAgent comment](https://github.com/hamstar/Wikimate/commit/f6983f739f17a9b05de1dbb695578f2321038484#diff-915f1e0f55c2400eb73093b0452dbac786cbcb0a04a0f5b0db0e4adaf5b81fbbR205) would be a more [effective reminder](https://github.com/hamstar/Wikimate/pull/107#discussion_r664447951) if it's also in the Usage section?
> 
> Definitely. And this PR seems like a good place to add that change :)

Added as separate commit 2045aa9.

USAGE.md
```diff
@@ -339,3 +348,20 @@ $useragent = $wiki->getUserAgent();
 $wiki->setUserAgent('Custom User Agent');
 $wiki->setUserAgent('Custom Prefix - ' . $useragent);
 ```
+
+In order to use a custom user agent for all requests in the session,
+call this method before invoking Wikimate::login().
```
waldyrious, line 353: Shouldn't this be formatted as code?
```suggestion
call this method before invoking `Wikimate::login()`.
```

waldyrious:
> Hmm, no original commit message did, but I can add it to both?

Oh, right, I was confused by github's weird habit of showing the PR title as the main title (and the html `<title>` when viewing a PR's commit, like [this](https://github.com/hamstar/Wikimate/pull/108/commits/8d0c91ff82c4fef96e3ed46a09434df064a0e64c). Anyway, yeah, please add the context to all commit messages so they can be read independently from the other ones.

> How does everything look now?

Should be ready after the commit messages are adjusted, and the inline comment I just added is addressed.

USAGE.md
Xymph: Naturally :roll_eyes: Thx.

Xymph:
> Should be ready after the commit messages are adjusted, and the inline comment I just added is addressed.

New round pushed.

waldyrious (approved):
LGTM, nice work!

PR #109, opened 2021-07-07
Add missing PHPDoc comments for properties, constants, and more

Xymph:
Running PHPDoc (v2.9.1) highlighted that most class properties and constants were lacking comments, and existing ones were incomplete. Also, parameters to the Wikimate constructor method weren't documented, and the main file header was misformatted using /// comments.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -15,22 +15,81 @@
 class Wikimate
 {
 	/**
-	 * @var  string  The current version number (conforms to http://semver.org/).
+	 * The current version number (conforms to Semantic Versioning)
+	 *
+	 * @var string
+	 * @link https://semver.org/
 	 */
 	const VERSION = '0.13.0';
 
+	/**
+	 * Identifier for CSRF token
+	 *
+	 * @var string
+	 * @link https://www.mediawiki.org/wiki/Special:MyLanguage/API:Tokens
+	 */
 	const TOKEN_DEFAULT = 'csrf';
-	const TOKEN_LOGIN   = 'login';
 
+	/**
+	 * Identifier for Login token
+	 *
+	 * @var string
+	 * @link https://www.mediawiki.org/wiki/Special:MyLanguage/API:Tokens
+	 */
+	const TOKEN_LOGIN = 'login';
+
+	/**
+	 * Base URL for API requests
+	 *
+	 * @var string
+	 * @link https://www.mediawiki.org/wiki/Special:MyLanguage/API:Main_page#Endpoint
+	 */
 	protected $api;
+
+	/**
+	 * Username for API requests
+	 *
+	 * @var string
+	 * @link https://www.mediawiki.org/wiki/Special:MyLanguage/API:Login#Method_1._login
+	 */
 	protected $username;
+
+	/**
+	 * Password for API requests
+	 *
+	 * @var string
+	 * @link https://www.mediawiki.org/wiki/Special:MyLanguage/API:Login#Method_1._login
+	 */
 	protected $password;
 
-	/** @var  Requests_Session */
+	/**
+	 * Session object for HTTP requests
+	 *
+	 * @var Requests_Session
+	 * @link https://requests.ryanmccue.info/
+	 */
 	protected $session;
+
+	/**
+	 * User agent string for Requests_Session
+	 *
+	 * @var string
+	 * @link https://requests.ryanmccue.info/docs/usage-advanced.html#session-handling
+	 */
 	protected $useragent;
 
-	protected $error     = null;
+	/**
+	 * Error array with API and Wikimate errors
+	 *
+	 * @var array|null
+	 */
+	protected $error = null;
+
+	/**
+	 * Whether to output debug logging
+	 *
+	 * @var boolean
+	 */
 	protected $debugMode = false;
 
 	/**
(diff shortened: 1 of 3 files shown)


Wikimate.php
```diff
@@ -1,41 +1,107 @@
...
-	protected $error     = null;
+	/**
+	 * Error array with API and Wikimate errors
+	 *
+	 * @var array|null
+	 */
+	protected $error = null;
+
+	/**
+	 * Should we output debug logging?
```
waldyrious, line 92: This wording could be confusing and seem like a TODO/FIXME note. How about this?
```suggestion
	 * Whether to output debug logging
```

Wikimate.php
```diff
@@ -495,23 +567,87 @@ public function getError()
 /**
  * Models a wiki article page that can have its text altered and retrieved.
  *
- * @author  Robert McLeod
+ * @author  Robert McLeod & Frans P. de Vries
```
waldyrious, line 570: Maybe update the copyright notice in the license file as well?

https://github.com/hamstar/Wikimate/blob/bc3214a56f8c21d48280a747e8aff69dad0810d8/LICENSE.md#L3

Wikimate.php
```diff
@@ -495,23 +567,87 @@ public function getError()
...
  * @since   December 2010
  */
 class WikiPage
 {
+	/**
+	 * Section list identifier for index
```
waldyrious, line 576: This wording is not clear to me. What do you mean by "for index"?

Wikimate.php
```diff
@@ -495,23 +567,87 @@ public function getError()
...
+
+	/**
+	 * Wikimate object for API requests
+	 *
+	 * @var Wikimate|null
+	 */
+	protected $wikimate = null;
+
+	/**
+	 * Does the page exist?
```
waldyrious, line 611: Same as [above](https://github.com/hamstar/Wikimate/pull/109/files#r665263159):

```suggestion
	 * Whether the page exists
```

(Or something more verbose, like "Indicator of whether the page exists")

Wikimate.php
```diff
@@ -1,41 +1,107 @@
...
  *
- * @author  Robert McLeod
+ * @author  Robert McLeod & Frans P. de Vries
  * @since   December 2010
  */
 class Wikimate
 {
 	/**
-	 * @var  string  The current version number (conforms to http://semver.org/).
+	 * The current version number (conforms to Semantic Versioning).
```
waldyrious, line 21: None of the other descriptions (in this PR) use punctuation, so it seems like this shouldn't either.

Wikimate.php
Xymph: Saw the 'should?' wording in Requests and elsewhere a few times, but 'whether' is fine too.

Wikimate.php
Xymph: Done.

Wikimate.php
Xymph: Dot dropped.

Wikimate.php
Xymph: I'll stick with the concise here :)

Wikimate.php
Xymph: I just knew that shortcut would boomerang back to me. :wink: Rephrased.

Xymph:
Reworked feedback in e7bc048. Btw, did you see my email on your gmail?

Wikimate.php
waldyrious: Much clearer now, thanks!

waldyrious:
> Btw, did you see my email on your gmail?

I just did, and responded.

By the way, could you add [this comment](https://github.com/hamstar/Wikimate/pull/109#commitcomment-53167942) in the [PR diff](https://github.com/hamstar/Wikimate/pull/109/files), rather than the [stand-alone commit page](https://github.com/hamstar/Wikimate/commit/e7bc0480a22ba721a4c2f07aad6312180623e8e2)? GitHub get a bit confused by comments added there, and they can't be marked as resolved. (If you do that, I'll copy my response as well, and then we can delete the comments from the commit page to clean things up.)

Wikimate.php
```diff
@@ -495,23 +567,80 @@ public function getError()
...
+	 */
 	const SECTIONLIST_BY_INDEX = 1;
+
+	/**
+	 * Use section names as keys in return array of {@link WikiPage::getAllSections()}
+	 *
+	 * @var integer
+	 */
 	const SECTIONLIST_BY_NAME = 2;
-	const SECTIONLIST_BY_NUMBER = 3;
```
Xymph, line 505: TBH, I don't know what the idea behind this third section-list key type could be. It has been in the code [since v0.4](https://github.com/hamstar/Wikimate/commit/8220b04d58b7de138724aef9e73084f7d2bece45#diff-915f1e0f55c2400eb73093b0452dbac786cbcb0a04a0f5b0db0e4adaf5b81fbbR119), but even then it wasn't used and falls into the [default clause](https://github.com/hamstar/Wikimate/commit/8220b04d58b7de138724aef9e73084f7d2bece45#diff-915f1e0f55c2400eb73093b0452dbac786cbcb0a04a0f5b0db0e4adaf5b81fbbR399) of the switch.

I propose to remove this constant altogether.

waldyrious: Yeah, it seems like it would be the same as the index. I agree with removing it as it isn't used. If eventually @hamstar is able to clarify the original intent, we can reintroduce it.

Xymph:
> By the way, could you add [this comment](https://github.com/hamstar/Wikimate/pull/109#commitcomment-53167942) in the [PR diff](https://github.com/hamstar/Wikimate/pull/109/files), rather than the [stand-alone commit page](https://github.com/hamstar/Wikimate/commit/e7bc0480a22ba721a4c2f07aad6312180623e8e2)? GitHub get a bit confused by comments added there, and they can't be marked as resolved. (If you do that, I'll copy my response as well, and then we can delete the comments from the commit page to clean things up.)

Copied, and original comment removed. The constant itself is also gone.

waldyrious (approved):
LGTM 👍

Issue #104, opened 2021-07-04
Roadmap


Xymph:
Don't know if this is a common approach, but a discussion ticket seems a suitable way to talk about short-term plans for Wikimate updates.

I would like to wrap up #102 asap and move forward, as work on four new PRs is already done and tested. They are, in order:
1. issue #103
1. add table of contents to USAGE doc
1. change API calls from PHP format to JSON
1. issue #92

Regarding 3: PHP format is [deprecated ](https://www.mediawiki.org/wiki/API:Data_formats#Output)and only maintained for now due to [its popularity](https://www.mediawiki.org/wiki/API:Data_formats#Additional_notes). New users should stick [to JSON](https://www.mediawiki.org/wiki/API:Etiquette#Data_formats), and some API calls [like opensearch](https://www.mediawiki.org/wiki/API:Opensearch) don't even support PHP format.

The other reason to switch to JSON is that my implementation for 4. moves all API requests into a single request() method, which can then also contain the decoding step (and debug logging of request and response) in one central place, instead of the current multiple unserialize-s in methods like query(), edit() and upload(). And such centralizing can only be done if all requests use the same JSON format, as it is the only one supported by all API calls.

A format switch would ideally be verified with unit testing of those Wikimate methods, but this is an area I have little experience in, and it would be time-consuming to set this up. My own application scripts continue to work fine with the changes in 3. and 4. though.

The easy way (also avoiding potential merging conflicts) is to implement these PRs in the repo in the above order, so I'd prefer to just create and merge one branch at a time (although 1-2 can be done in parallel with 3-4). But that brings me back to the (slow) progress of #102, and @waldyrious 's availability here. I certainly don't want to you as a punching bag my friend, and my afterthought commits in #102 demonstrate that reviews/feedback on my stuff is no luxury :wink: , but I would prefer not to stay stalled for long(er) times either.

Please share any thoughts on the above.

As for the longer term, I intend to look into existing issues again, and possibly explore coverage of more API features, but have no definite ideas/plans yet.

Xymph:
I just ran phpdoc on Wikimate.php for the first time, and it pointed out some missing summaries, mostly for properties. So another planned PR is to add the absent info.

waldyrious:
> Don't know if this is a common approach, but a discussion ticket seems a suitable way to talk about short-term plans for Wikimate updates.

Yeah, I agree that roadmap issues like this sound like a great way to plan out work.

[...]

PR #111, opened 2021-07-07
Change API requests from deprecated PHP format to JSON format

Xymph:
PHP format is [deprecated](https://www.mediawiki.org/wiki/API:Data_formats#Output) and only maintained for now due to [its popularity](https://www.mediawiki.org/wiki/API:Data_formats#Additional_notes). New users should stick [to JSON](https://www.mediawiki.org/wiki/API:Etiquette#Data_formats), and some API calls [like opensearch](https://www.mediawiki.org/wiki/API:Opensearch) don't even support PHP format. See PR #104 for further discussion.

The Wikimate methods return nested associative arrays, which were produced by [unserialize](https://www.php.net/manual/en/function.unserialize) from the PHP format. Fortunately, while [json_decode](https://www.php.net/manual/en/function.json-decode) normally returns a stdClass object, setting the $associative flag will produce the same arrays.

Ideally this switch would be unit-tested with before/after comparisons, but that is beyond my expertise and available time. I have however run my client scripts with the switched format for a while and not encountered any issues.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -163,7 +163,7 @@ class Wikimate
 			return false;
 		}
 
-		$tokenResult = json_decode($response->body);
+		$tokenResult = json_decode($response->body, true);
 		// Check if we got a JSON result
 		if ($tokenResult === null) {
 			$this->error = array();
@@ -179,9 +179,9 @@ class Wikimate
 		}
 
 		if ($type == self::TOKEN_LOGIN) {
-			return $tokenResult->query->tokens->logintoken;
+			return $tokenResult['query']['tokens']['logintoken'];
 		} else {
-			return $tokenResult->query->tokens->csrftoken;
+			return $tokenResult['query']['tokens']['csrftoken'];
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Wikimate.php
```diff
@@ -163,7 +163,7 @@ protected function token($type = self::TOKEN_DEFAULT)
 			return false;
 		}
 
-		$tokenResult = json_decode($response->body);
+		$tokenResult = json_decode($response->body, true);
```
Xymph, line 166: Strictly speaking, this change (and the similar one in login() ) isn't needed as the Result variables aren't used outside these methods. But mixing two different decoding styles in the class could be a bit confusing. So now they all use the associative array style.

waldyrious:
Overall this LGTM — I don't have a setup to test it myself, but I trust that your tests are sufficient to provide some assurance that things are working as expected. 

I'm thinking that this would normally require a major version, since we're changing the tool's interface. However, we're still using a 0.x "development" version scheme, so that wouldn't apply in this case. Still, perhaps this change would be a good moment to mark 1.0 (unless you're planning further backward-incompatible changes). WDYT?

waldyrious (approved):

PR #114, opened 2021-07-09
Remove null returns from destructors & fix PHPDoc comments

Xymph:
Destructors are of [type void](https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.destructor), so we shouldn't return null values.
Actually noticed this because PHPDoc complained about the invalid "&lt;type&gt;" syntax. A good occasion to also fix inlined internal/external links (at-link apparently doesn't work so well inlined).

Edit: the header bars didn't look good in the generated test tree, and the at-since tags were lacking versions -- also fixed.


waldyrious (approved):
All of this LGTM. Happy to merge whenever you're ready :)

PR #121, opened 2021-08-24
Fix user agent format; add refs & changelog entry

Xymph:
[MediaWiki etiquette](https://www.mediawiki.org/wiki/API:Etiquette#The_User-Agent_header) prescribes the user agent string to include "client/version", where Wikimate used a space instead of the slash. This PR addresses that and adds reference links to the property and methods.


waldyrious (approved):
LGTM!

PR #122, opened 2021-08-24
Update Wikimate::token() to remember CSRF token (fixes #115)

### LINKED ISSUE #115: Fix repeated CSRF token fetching
PR #100 introduced the `Wikimate::token()` method, invoked by every data-modifying API call. But a [CSRF token is valid](https://www.mediawiki.org/wiki/API:Edit#Additional_notes) for an entire login session. `token()` requests a new one every time, causing unnecessary repetitive API calls.

The method should store the first requested CSRF token and return that upon further calls instead of making more API requests. It should be discarded in a (future) logout method.

Xymph:
See issue #115. Storing and re-using the token reduces the number of API calls when invoking multiple edit/delete/etc actions in a session, as is typical bot behavior.

The error value of `token()` is also changed to `null` (`false` should only be used by boolean methods). Since `token()` is an internal method, I don't consider this an API-breaking change.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -248,7 +248,7 @@ class Wikimate
 	 * for more information.
 	 *
 	 * @param   string  $type  The token type
-	 * @return  string         The requested token
+	 * @return  mixed          The requested token (string), or null if error
 	 */
 	protected function token($type = self::TOKEN_DEFAULT)
 	{
@@ -256,7 +256,7 @@ class Wikimate
 		if ($type != self::TOKEN_DEFAULT && $type != self::TOKEN_LOGIN) {
 			$this->error = array();
 			$this->error['token'] = 'The API does not support the token type';
-			return false;
+			return null;
 		}
 
 		// Check for existing CSRF token for this login session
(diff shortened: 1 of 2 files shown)


Wikimate.php
```diff
@@ -226,22 +235,30 @@ private function request($data, $headers = array(), $post = false)
...
+	 * upon further such requests, instead of making another API call.
 	 * For now this method, in Wikimate tradition, is kept simple and supports
 	 * only the two token types needed elsewhere in the library.  It also
 	 * doesn't support the option to request multiple tokens at once.
 	 * See {@see https://www.mediawiki.org/wiki/Special:MyLanguage/API:Tokens}
 	 * for more information.
 	 *
 	 * @param   string  $type  The token type
-	 * @return  string         The requested token
+	 * @return  mixed          The requested token (string), or null if error
```
waldyrious, line 248: Just a small nitpick: "null if error" → "null in case of error".

waldyrious (approved):
LGTM, but not merging immediately in case you want to address the nitpick comment I left inline.

Wikimate.php
Xymph: Hmm, there are five more places using the same phrase. Either all are changed, or kept as is. I think it's fine as is.

Xymph:
Not really, but I think I'll need to rebase/push for the merge conflict first, right?

Wikimate.php
waldyrious: Yeah, makes sense. Let's leave it as is.

waldyrious:
I think it's nice to rebase even when there aren't conflicts, to avoid crossing branches and provide a [semi-linear git graph](https://github.com/isaacs/github/issues/1017). :)

PR #123, opened 2021-08-25
Add methods WikiFile::revert() and Wikimate::filerevert()

Xymph:
Covers the API [action Filerevert](https://www.mediawiki.org/wiki/Special:MyLanguage/API:Filerevert). Not much more to say about it.


waldyrious (approved):
LGTM :+1:

Issue #116, opened 2021-07-11
Fix error return values in WikiPage::getSection and setText


Xymph:
Commit aeaa486 introduced checks in `getSection()` for undefined sections that return false if not found. This is inconsistent with other methods (e.g. `getText()`, `Wikimate::download()` and many Wikifile methods) which return null to indicate error (and otherwise a string or array). Only boolean methods should return false upon error, I think.

Also, recent commit 0a00740 [returns a null](https://github.com/hamstar/Wikimate/commit/0a00740f2413e6966ade51ec9edc3bac762af4f6#diff-915f1e0f55c2400eb73093b0452dbac786cbcb0a04a0f5b0db0e4adaf5b81fbbR877) upon error in boolean method `setText()`, which should be false.

The getSection change is API breaking, so it should go in just before we bump to v1.0.0.

Issue #89, opened 2018-07-25
Feature request: Basic auth authentication


finalbeta:
Would bring compatibility with https://www.mediawiki.org/wiki/Extension:Auth_remoteuser

> This allows integration with the web server's built-in authentication system (for example via the REMOTE_USER environment variable, which is set through HTTP-Auth, LDAP, CAS, PAM, etc.) or any other type of external authentication (SSL client auth, user accounts provided by different forum software, etc.).

Issue #87, opened 2017-12-02
Feature request: OAuth authorization


ms609:
I can't see a way to use [OAuth authentication](https://www.mediawiki.org/wiki/OAuth/Owner-only_consumers) instead of logging in using a username and password; this would be a handy feature.

PR #124, opened 2021-08-25
Add method Wikimate::logout()

Xymph:
Covers the API [action Logout](https://www.mediawiki.org/wiki/Special:MyLanguage/API:Logout) and discards CSRF token (per #115).

The error code value is 'auth' because it doesn't make much sense to define a code 'logout' that's unlikely ever checked. Instead it would be logical to change the 'login' code into 'auth' as well (possibly also useful for #87 and #89), but that would be an small API-breaking change (similar to #116), so has to be delayed til v1.0.0. Hence the 'auth' entry is added to USAGE.md but the 'login' entry not yet removed.

A logout() example in USAGE.md would be too trivial so I didn't bother.
USAGE.md also includes a user agent fix I missed in #121.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -341,7 +341,7 @@ Both methods return an array of the MediaWiki API result.
 API requests are made over HTTP with a user agent string to identify
 the client to the server. By default the user agent is formatted as:
 
-`Wikimate <VERSION> (https://github.com/hamstar/Wikimate)`
+`Wikimate/<VERSION> (https://github.com/hamstar/Wikimate)`
 
 The string can be retrieved and customized via:
 
@@ -363,7 +363,8 @@ print_r($page->getError());
 ```
 
 For MediaWiki API errors, the array contains the 'code' and 'info' key/value pairs [defined by the API](https://www.mediawiki.org/wiki/Special:MyLanguage/API:Errors_and_warnings#Errors).  For other errors, the following key/value pairs are returned:
-* 'login' for Wikimate authentication problems
+* 'login' for Wikimate authentication problems <!-- TODO: remove after changing login() to use the 'auth' code too -->
+* 'auth' for Wikimate authentication problems
 * 'token' for Wikimate token problems
 * 'page' for WikiPage errors
 * 'file' for WikiFile errors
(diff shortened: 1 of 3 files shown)


Wikimate.php
```diff
@@ -374,6 +374,56 @@ public function login($username, $password, $domain = null)
...
+		}
+
+		if ($this->debugMode) {
+			echo "Logout request:\n";
+			print_r($details);
+			echo "Logout response:\n";
+			print_r($logoutResult);
+		}
+
+		// Discard CSRF token for this login session
```
waldyrious, line 422: Maybe add a note [here](https://github.com/hamstar/Wikimate/blob/3172a7863545f7020cc08fe55192afe83e971c7c/Wikimate.php#L125) and/or [here](https://github.com/hamstar/Wikimate/blob/3172a7863545f7020cc08fe55192afe83e971c7c/Wikimate.php#L242-L243) about when the CSRF token gets discarded?

USAGE.md
```diff
@@ -364,6 +364,7 @@ print_r($page->getError());
 
 For MediaWiki API errors, the array contains the 'code' and 'info' key/value pairs [defined by the API](https://www.mediawiki.org/wiki/Special:MyLanguage/API:Errors_and_warnings#Errors).  For other errors, the following key/value pairs are returned:
 * 'login' for Wikimate authentication problems
+* 'auth' for Wikimate authentication problems
```
waldyrious, line 367: Perhaps we could add an HTML comment (`<!-- -->`) in the line above to explain the temporary duplication.

waldyrious (commented):
Left a couple suggestions inline for the documentation. The code itself LGTM.

USAGE.md
waldyrious: (A "TODO: remove when..." note would work, btw.)

Wikimate.php
Xymph: Good idea 1, done.

USAGE.md
Xymph: Good idea 2, done.

waldyrious (approved):
LGTM!

PR #120, opened 2021-08-23
Prepare for v0.14.0 release

Xymph:
Release summary:

"This release adds support for the maxlag parameter and for getting/setting the user agent, changes API calls from deprecated PHP format to JSON format, and improves PHPDoc commentary and USAGE.md.
It requires MediaWiki v1.27 or newer. For more details on these changes, including links to the corresponding pull requests, please check the Changelog."

PR #125, opened 2021-08-25
Add post-release update steps to GOVERNANCE.md

Xymph:
As [proposed here](https://github.com/hamstar/Wikimate/pull/120#issuecomment-904957481).

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -53,3 +53,7 @@ The version tag and the title of the release notes should be in the format v1.2.
 to determine which part of the version number to increase).
 The body of the release notes should be a summary of the contents
 of the relevant section in `CHANGELOG.md`.
+
+Finally, update the Wikimate entry on [Packagist](https://packagist.org/packages/hamstar/wikimate) (via the "Update Now" link in the right sidebar)
+and the corresponding row in the [PHP libraries table](https://www.mediawiki.org/wiki/API:Client_code/All#PHP) on MediaWiki.org,
+along with Wikipedia's [PHP bot framework table](https://en.wikipedia.org/wiki/Wikipedia:PHP_bot_framework_table) if applicable.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Since v0.10.0 this project adheres to [Semantic Versioning](http://semver.org/) 
 
 * New methods `WikiFile::revert()` and `Wikimate::filerevert()` ([#123])
 * New method `Wikimate::logout()` ([#124])
+* Added post-release update steps to GOVERNANCE.md ([#125])
 
 #### Changed
 
@@ -153,3 +154,4 @@ Since v0.10.0 this project adheres to [Semantic Versioning](http://semver.org/) 
 [#122]: https://github.com/hamstar/Wikimate/pull/122
 [#123]: https://github.com/hamstar/Wikimate/pull/123
 [#124]: https://github.com/hamstar/Wikimate/pull/124
+[#125]: https://github.com/hamstar/Wikimate/pull/125


GOVERNANCE.md
```diff
@@ -53,3 +53,5 @@ The version tag and the title of the release notes should be in the format v1.2.
 to determine which part of the version number to increase).
 The body of the release notes should be a summary of the contents
 of the relevant section in `CHANGELOG.md`.
+
+Finally, the Wikimate entry on [Packagist](https://packagist.org/packages/hamstar/wikimate) and the corresponding row in the [PHP libraries table](https://www.mediawiki.org/wiki/API:Client_code/All#PHP) on MediaWiki.org should be updated, along with Wikipedia's [PHP bot framework table](https://en.wikipedia.org/wiki/Wikipedia:PHP_bot_framework_table) if applicable.
```
waldyrious, line 57: Please break the lines using the [semantic linefeeds](https://rhodesmill.org/brandon/2012/one-sentence-per-line/) pattern, for consistency with the rest of the document. For example:

```
Finally, update the Wikimate entry on [Packagist](https://packagist.org/packages/hamstar/wikimate)
and the corresponding row in the [PHP libraries table](https://www.mediawiki.org/wiki/API:Client_code/All#PHP) on MediaWiki.org,
along with Wikipedia's [PHP bot framework table](https://en.wikipedia.org/wiki/Wikipedia:PHP_bot_framework_table) if applicable.
```

On a more content-related note, we probably should describe how to update the Packagist entry (i.e., log in with one's GitHub account and click the "Update Now" link at the bottom of the right sidebar). WDYT?

waldyrious (commented):
Thanks for doing this! :heart:
I left two minor comments inline.

GOVERNANCE.md
Xymph: Oops, yes, I forgot. Lines wrapped.

As far as instructions how to use external sites, I think that doesn't belong here (or anywhere in Wikimate). Logging in can also be done via a Packagist account, and even then it's trivial to update the package although the link is a bit obscure indeed. We don't (and shouldn't) describe how to edit those wiki pages either. Or do you strongly feel we should?

waldyrious: I don't think it's necessary to explain how to edit wiki pages on a mediawiki wiki, given the scope of this project. The behavior of Packagist may be a little harder to figure out, though. How about just mentioning the need to log in before updating the package? Something simple, like:

```
Finally, manually update the Wikimate entry on [Packagist](https://packagist.org/packages/hamstar/wikimate) (remember to log in first)
and edit the corresponding row in the [PHP libraries table](https://www.mediawiki.org/wiki/API:Client_code/All#PHP) on MediaWiki.org,
along with Wikipedia's [PHP bot framework table](https://en.wikipedia.org/wiki/Wikipedia:PHP_bot_framework_table) if applicable.
```

Rendered:

> Finally, manually update the Wikimate entry on [Packagist](https://packagist.org/packages/hamstar/wikimate) (remember to log in first) and edit the corresponding row in the [PHP libraries table](https://www.mediawiki.org/wiki/API:Client_code/All#PHP) on MediaWiki.org, along with Wikipedia's [PHP bot framework table](https://en.wikipedia.org/wiki/Wikipedia:PHP_bot_framework_table) if applicable.

Xymph: Logging in anywhere is the no-brainer part. Then I prefer to point out the link (if it's missing, log in :wink: ). That work for you?

waldyrious: Yeah, good idea — at least that indicates what one should be looking for :)

waldyrious (approved):
Nice one! LGTM.

GOVERNANCE.md
waldyrious: For future reference, #135 introduces the semantic line breaks recommendation explicitly into the project docs.

PR #130, opened 2021-08-26
Update guidelines for commits and pull requests

waldyrious:
- Make the first sentence of each entry bold, for readability and consistency
- Add a new entry regarding merge strategy (to achieve a semi-linear git history, as suggested [here](https://github.com/hamstar/Wikimate/pull/122#issuecomment-905444038))
- Slightly reword a couple passages for clarity

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ and [Keep a Changelog](http://keepachangelog.com/).
 
 ### Upcoming version
 
+#### Changed
+
+* Added additional context to README.md ([#127])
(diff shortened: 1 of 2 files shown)


waldyrious:
> - Add a new entry regarding merge strategy (to achieve a semi-linear git history, as suggested [here](https://github.com/hamstar/Wikimate/pull/122#issuecomment-905444038))

For future reference, we would ideally configure the repository to only allow "create a merge commit" as a merge strategy for PRs, but that would require admin access, which so to this day only @hamstar has. Hopefully we will eventually be able to improve the project's bus factor and have more admins.

PR #112, opened 2021-07-07
Support for the maxlag parameter (with retries) in API requests (fixes #92)

Xymph:
This parameter insures that API requests can be dropped after the specified number of seconds by a replicated wiki server cluster (such as Wikipedia) if it is strained too much. If a maxlag response is returned, Wikimate delays (sleeps) for a given number of seconds before retrying the request. It can do this indefinitely, or for a limited number of retries. See the documentation at https://www.mediawiki.org/wiki/Manual:Maxlag_parameter

For the implementation I looked around at other PHP MediaWiki libraries for inspiration, and in the [bot functions table](https://en.wikipedia.org/wiki/Wikipedia:PHP_bot_functions) (linked from [here](https://en.wikipedia.org/wiki/Help:Creating_a_bot#PHP)) found an abandoned library called [Pillar](https://sourceforge.net/projects/pillar/). Here the [request class](https://sourceforge.net/p/pillar/code/HEAD/tree/trunk/class.mwrequest.php) offered a useful example of handling the lag.

PR #134, opened 2021-08-29
Add usage documentation about maximum lag and retries

Xymph:
In all the work and discussion regarding #112, I took a shortcut in not documenting the feature, and my trusty reviewer didn't remind me about it either. :wink: 

The text already uses `WikimateException` which doesn't exist yet, but will be defined in my next PR (after this and #133 are merged).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ and [Keep a Changelog](http://keepachangelog.com/).
 
 ### Upcoming version
 
+#### Added
+
+* Usage documentation about maximum lag and retries ([#134])
+
 #### Changed
 
 * Added additional context to README.md ([#127])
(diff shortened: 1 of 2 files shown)


waldyrious (approved):
LGTM :) I'll wait until this is rebased to merge it.

waldyrious:
(I do have to mention that I feel a little uneasy with the time-traveling aspect of an earlier commit mentioning a concept that will be introduced in a later commit, but hey, nothing wrong with leaving a little mystery to the future code archeologists :P)

Xymph:
> (I do have to mention that I feel a little uneasy with the time-traveling aspect of an earlier commit mentioning a concept that will be introduced in a later commit, but hey, nothing wrong with leaving a little mystery to the future code archeologists :P)

No offense intended, but would you have felt equally uneasy if I hadn't pointed it out explicitly?
Anyway, thanks for working with me, even into the future. :wink:

waldyrious:
> No offense intended, but would you have felt equally uneasy if I hadn't pointed it out explicitly?

[Good question](https://en.wikipedia.org/wiki/If_a_tree_falls_in_a_forest)! I would likely not have noticed it — or rather, I'd assume it already existed in the codebase, which I don't know nearly as intimately as you do 🙂.

PR #136, opened 2021-08-29
Add WikimateException and centralize API error checks in request() 

Xymph:
This implements the exception handling portion of #113, where the reasoning for this approach is discussed. Debug logging changes will go into a separate PR, which will close that issue.

A commit with documentation/changelog edits follows after #135 is wrapped up and merged. (Hmm, I guess this should have been a draft PR. Can it still be changed?)

PR #139, opened 2021-09-01
Centralize API debug logging in Wikimate::request() (fixes #113)

### LINKED ISSUE #113: Centralize API request/response debug logging and response decoding
PR #112 introduced the `Wikimate::request()` method, invoked for all API calls. The latter still check the response body for some error situations -- the [doc-page test](https://github.com/hamstar/Wikimate/blob/master/Wikimate.php#L159) and the [decoding result](https://github.com/hamstar/Wikimate/blob/master/Wikimate.php#L167) -- and perform the JSON decode. The error checks happen only in `token()` and `login()` anyway. These could be moved into `request()`, but it is also worth considering not to return an error array but throw an exception. I have to ponder this a bit longer. Any thoughts, @waldyrious ?

Secondly, the debug logging of the request array and decoded response array can also be centralized in `request()`.

Xymph:
See discussion in issue #113. This is the "other half" of the centralization steps started with #136. Combined they reduce the overall length of `Wikimake.php` by some 50 lines, and significantly streamline all methods that call `request()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,11 +63,6 @@ else {
 
 This example uses echo statements to output any potential errors.
 You should get a meaningful error message if the authentication fails.
-
-Instead of using echo statements, you can enable/disable debugging
-with the `$wiki->debugMode($boolean)` method.
-Currently only output from the logon process is printed for debugging.
-
 Assuming you were able to log in, you're now ready to fully use the API.
 
 See [USAGE.md](USAGE.md) for detailed example code to perform common tasks.
--- a/USAGE.md
+++ b/USAGE.md
@@ -15,6 +15,7 @@
   - [Customizing the user agent](#customizing-the-user-agent)
   - [Maximum lag and retries](#maximum-lag-and-retries)
   - [Handling errors and exceptions](#handling-errors-and-exceptions)
+  - [Debug logging](#debug-logging)
 
 ## Introduction
 
@@ -44,10 +45,6 @@ else {
 
 This example uses echo statements to output any potential errors.
 You should get a meaningful error message if the authentication fails.
-
-Instead of using echo statements, you can enable/disable debugging
-with the `$wiki->debugMode($boolean)` method.
-Currently only output from the logon process is printed for debugging.
 
 Assuming you were able to log in, you're now ready to fully use the API.
 The next sections provide example code for several common tasks.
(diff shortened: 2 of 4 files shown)


USAGE.md
```diff
@@ -47,7 +47,8 @@ You should get a meaningful error message if the authentication fails.
 
 Instead of using echo statements, you can enable/disable debugging
 with the `$wiki->debugMode($boolean)` method.
-Currently only output from the logon process is printed for debugging.
+Debug logging is printed to [standard output](https://en.wikipedia.org/wiki/Standard_output)
+and includes API requests/responses as well as retry messages for lag errors.
```
waldyrious, line 51:
```suggestion
and includes API requests/responses as well as retry messages for [lag errors](#maximum-lag-and-retries).
```

README.md
```diff
@@ -66,7 +66,8 @@ You should get a meaningful error message if the authentication fails.
 
 Instead of using echo statements, you can enable/disable debugging
 with the `$wiki->debugMode($boolean)` method.
-Currently only output from the logon process is printed for debugging.
+Debug logging is printed to [standard output](https://en.wikipedia.org/wiki/Standard_output)
+and includes API requests/responses as well as retry messages for lag errors.
```
waldyrious, line 70: I'm thinking that it may make sense to remove these two paragraphs from the README, in the interests of keeping the content simple (à la hello world) and reducing duplication with `USAGE.md`. Or do you think this is important information to convey for first-time users?

USAGE.md
waldyrious: Also, what do you think about moving this content to a dedicated section under "Other stuff", perhaps right before the "Handling errors and exceptions" section?

waldyrious (commented):
Nice changes overall — loving the code reduction and consolidation :)

I've left a couple suggestions inline, but I'd be happy to merge this as-is if you disagree with them.

USAGE.md
Xymph: Good idea, done (after it).

README.md
waldyrious: I'll assume that you deliberately chose to let the "This example uses echo statements..." part stay. I can see why that might make sense, but I'm just leaving a note for the record, since my original proposal above was to remove it as well.

waldyrious (approved):
LGTM!

README.md
Xymph: That was somewhat slightly deliberate, yeah. :smile_cat:

PR #140, opened 2021-09-02
Rewrite installation and invocation instructions using Composer

Xymph:
This PR was prompted by [this observation](https://github.com/hamstar/Wikimate/issues/104#issuecomment-905916260). Some of the text was copied from the [AddWiki docs](https://addwiki.github.io/v2/mediawiki-api-base/installation.html), I hope that is alright.

@samwilson, is this what you had in mind?

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,27 +23,46 @@ you will need to have PHP installed to run it.
 Install it with your preferred package management tool
 (for example, on Ubuntu Linux you can run: `sudo apt-get install php`)
 
+The recommended way to install this library is with Composer.
+Composer is a dependency management tool for PHP
+that allows you to declare the dependencies your project needs
+and installs them into your project.
+
 Install Composer by following the instructions [here](https://getcomposer.org/doc/00-intro.md).
 
-Then, download Wikimate, and initialise it by running `composer install`
-(or `composer.bat install` if you're on Windows).
+Then, run the following command in your project's folder
+to download Wikimate and initialise it:
 
-To use Wikimate within another project, you can add it as a composer dependency
-by adding the following to your `composer.json` file:
+```sh
+composer require hamstar/Wikimate
+```
+
+(or `composer.bat require hamstar/Wikimate` if you're on Windows).
+
+To use Wikimate within another project, you can add it as a Composer dependency
+by adding the following to your existing `composer.json` file:
 
 ```json
-"hamstar/Wikimate": "0.15.0"
+{
+    "require": {
+        "hamstar/Wikimate": "^0.15"
+    }
+}
 ```
+
+You can find out more on how to install Composer,
+configure autoloading, and other best-practices for defining dependencies
+at [getcomposer.org](https://getcomposer.org).
 
 ## Usage
 
-In your script file (e.g. `index.php`), include Wikimate's `globals.php` file,
+In your script file (e.g. `index.php`), include the project's `autoload.php` file,
 and create a new `Wikimate` object with the target wiki's API address.
 Then provide a username and password to Wikimate's `login` method,
 to log in to that wiki.
 
 ```php
-include 'globals.php';
+require __DIR__.'/vendor/autoload.php';
 
 $api_url = 'http://example.com/api.php';
 $username = 'bot';
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and [Keep a Changelog](http://keepachangelog.com/).
 
 - Centralized API communication checks in `WikiPage::request()` ([#136])
 - Centralized debug logging of API requests/responses in `WikiPage::request()` ([#139])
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -23,27 +23,45 @@ you will need to have PHP installed to run it.
...
+The recommended way to install this library is with Composer.
+Composer is a dependency management tool for PHP
+that allows you to declare the dependencies your project needs
+and installs them into your project.
+
 Install Composer by following the instructions [here](https://getcomposer.org/doc/00-intro.md).
 
-Then, download Wikimate, and initialise it by running `composer install`
-(or `composer.bat install` if you're on Windows).
+Then, download Wikimate and initialise it by running:
```
waldyrious, line 33: Perhaps in addition to the comma removal, reordering the sentence can make it clearer that the downloading is not a step to be performed manually:

```suggestion
Then, run the following command to download Wikimate and initialise it:
```

Also, it might be a nice touch to specify where this command should be run (e.g. "run the following command in your project's folder").

README.md
```diff
@@ -23,27 +23,45 @@ you will need to have PHP installed to run it.
...
 ```
 
+You can find out more on how to install Composer,
+configure autoloading, and other best-practices for defining dependencies
+at [getcomposer.org](https://getcomposer.org).
+
 ## Usage
 
-In your script file (e.g. `index.php`), include Wikimate's `globals.php` file,
+In your script file (e.g. `index.php`), include the project's `autoload.php` file,
```
waldyrious, line 58: Why is "the project's `autoload.php`" better than "Wikimate's `autoload.php`"? Am I missing something?

waldyrious (commented):
Nice! I left a couple suggestions inline, let me know what you think.

README.md
Xymph: A project has only one `autoload.php`. If Wikimate is "merely" a dependency in a bigger project, specifying "Wikimate's" file would be confusing. It is a library after all; as developers we're the only ones using Wikimate as our main project.

README.md
Xymph: Good idea, done.

README.md
waldyrious: Ah, gotcha. Thanks for clarifying :)

waldyrious (approved):
LGTM :) Let's see what @samwilson has to say.

README.md
```diff
@@ -23,27 +23,46 @@ you will need to have PHP installed to run it.
...
-To use Wikimate within another project, you can add it as a composer dependency
-by adding the following to your `composer.json` file:
+To use Wikimate within another project, you can add it as a Composer dependency
+by adding the following to your existing `composer.json` file:
 
 ```json
-"hamstar/Wikimate": "0.15.0"
+{
+    "require": {
+        "hamstar/Wikimate": "^0.15.0"
```
samwilson, line 48: It doesn't matter massively before 1.0, but I'd suggest writing this as the slightly more lenient `^0.15`. This is because (once the trailing zero becomes a 1 or whatever), if a project depends on two libraries that both depend on Wikimate, there's less chance of version conflicts. This doesn't really matter before 1.0, and anyway it's probably not that likely for multiple libraries to depend on Wikimate, so feel free to ignore this comment! :-)

samwilson (approved):
Looks good to me, reads well. I only have a tiny comment about the version number, and I was going to say that something should be added to the release process — but it's already mentioned in `GOVERNANCE.md`. :-)

Xymph (commented):
The prerelease example PR #81 in `GOVERNANCE.md` is a bit old, includes some discussion clutter, and doesn't show how to propose a Release Summary. Hence the change to short-and-clean #126.

waldyrious:
I'll merge in a bit, but first will change the title of the PR to be in the imperative mood, like the commit messages (and unlike the `CHANGELOG.md` file).

README.md
Xymph: > It doesn't matter massively before 1.0, but I'd suggest writing this as the slightly more lenient `^0.15`. This is because (once the trailing zero becomes a 1 or whatever), if a project depends on two libraries that both depend on Wikimate, there's less chance of version conflicts. This doesn't really matter before 1.0, and anyway it's probably not that likely for multiple libraries to depend on Wikimate, so feel free to ignore this comment! :-)

Thanks, I made the change.
But I am sort of wondering if what you wrote applies more to `~1.0` rather than `^1.0` (1.0.0 will indeed be our next release).

Per [the documentation](https://getcomposer.org/doc/articles/versions.md#next-significant-release-operators) my understanding is that `^1.0.0` will allow ">=1.0.0 <2.0.0" and is thus very lenient, while `~1.0.0` allows only ">1.0.0 <1.1.0" and thus indeed increases the risk of a version conflict. But I'm new to this, so open to enlightenment. :)

PR #142, opened 2021-10-16
Apply code formatting rules from PSR-2 and 12

Xymph:
[PSR](https://www.php-fig.org/psr/)-1, 2 and 12 provide code formatting rules, [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) (command `phpcs`) checks whether code adheres to each standard.

PSR-1 rule 3 defines that each class should be in a separate file, all in a single namespace. I'm not ready to make that change yet, but we should when adding another class.

PSR-2 rule 2.4 defines indenting to use 4 spaces, not tabs; 2.3 that lines should be soft-limited to 120 characters; 4.1 that class open/close braces should be on their own lines; 5.1 that `else if` should be written `elseif`.

PSR-12 rule 6.2 defines that binary operators should be surrounded by at least one space.

This PR ensures Wikimate meets the above rules. Besides the class separation, one other `phpcs` warning remains: "Visibility must be declared on all constants if your project supports PHP 7.1 or later". But since we support 5.3+, visibility keywords cannot be added.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -741,7 +741,9 @@ class Wikimate
  * @since   1.0.0  August 2021
  * @link    https://www.php.net/manual/en/class.runtimeexception.php
  */
-class WikimateException extends RuntimeException {}
+class WikimateException extends RuntimeException
+{
+}
 
 
 /**
@@ -1068,8 +1070,7 @@ class WikiPage
                         // Set the length of this section
                         $this->sections->byIndex[$currIndex]['length'] = $length;
                         $this->sections->byName[$currName]['length']   = $length;
-                    }
-                    else {
+                    } else {
                         // Set the length of last section
                         $this->sections->byIndex[$currIndex]['length'] = strlen($this->text) - $currOffset;
                         $this->sections->byName[$currName]['length']   = strlen($this->text) - $currOffset;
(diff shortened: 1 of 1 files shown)


Wikimate.php
```diff
@@ -741,7 +741,9 @@ public function getError()
  * @since   1.0.0  August 2021
  * @link    https://www.php.net/manual/en/class.runtimeexception.php
  */
-class WikimateException extends RuntimeException {}
+class WikimateException extends RuntimeException
+{
+}
```
waldyrious, line 747: Ugh. I wish the recommendations included an exception (heh) for the empty body case.

CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and [Keep a Changelog](http://keepachangelog.com/).
 
 ## Upcoming version
 
-No changes yet.
+### Changed
+
+- Applied code formatting rules from [PSR](https://www.php-fig.org/psr/) 2 and 12 ([#142])
```
waldyrious, line 12: Considering that [PSR-2](https://www.php-fig.org/psr/psr-2/) was deprecated in favor of [PSR-12](https://www.php-fig.org/psr/psr-12/) (which appears to be a superset of the former), shouldn't we mention only the latter?

Xymph: Fair point, I'll hitch-hike that into #143.

PR #143, opened 2021-10-29
Resolve static analysis warnings reported by PHPStan & PHPMD

Xymph:
[PHPStan](https://phpstan.org/) points out potential code problems via static analysis. Running `vendor/bin/phpstan analyse -c phpstan.neon` with this config file:
```
parameters:
	level: 6
	paths:
		- Wikimate.php
	checkMissingIterableValueType: false
```
listed the following warnings at [this level](https://phpstan.org/user-guide/rule-levels) (line numbers prior to the commit):
```
 ------ ---------------------------------------------------------------------- 
  Line   Wikimate.php                                                          
 ------ ---------------------------------------------------------------------- 
  146    Access to an undefined property Wikimate::$headers.                   
  147    Access to an undefined property Wikimate::$data.                      
  148    Access to an undefined property Wikimate::$options.                   
  164    Access to an undefined property Wikimate::$data.                      
  164    Access to an undefined property Wikimate::$headers.                   
  164    Access to an undefined property Wikimate::$options.                   
  587    Method Wikimate::edit() should return array but returns false.        
  611    Method Wikimate::delete() should return array but returns false.      
  659    Method Wikimate::upload() should return array but returns false.      
  710    Method Wikimate::filerevert() should return array but returns false.  
  907    Method WikiPage::destroy() has no return typehint specified.          
  1175   Unreachable statement - code above always terminates.                 
  1496   Method WikiFile::destroy() has no return typehint specified.          
 ------ ---------------------------------------------------------------------- 
```
The first commit addresses all these, little point in splitting them up.

The `checkMissingIterableValueType` config flag disables a [frequent warning](https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type) I don't want to bother with.

A handful more warnings remain (line numbers after the commit):
```
  186    Access to an undefined property Requests_Session::$useragent.  
  511    Access to an undefined property Requests_Session::$useragent.  
  1158   Variable $offset might not be defined.                         
  1163   Variable $depth might not be defined.                          
  1164   Variable $length might not be defined.                         
  1178   Variable $length might not be defined.                         
  1178   Variable $offset might not be defined.                         
  1181   Variable $offset might not be defined.      
```                   
The first pair cannot be resolved because the structure of the `$session` object returned by Requests_Session is unknown, and there is no setter method to replace the simple assignment.
The variable warnings are not worth the trouble at this time either, IMHO.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ and [Keep a Changelog](http://keepachangelog.com/).
 
 ### Changed
 
-- Applied code formatting rules from [PSR](https://www.php-fig.org/psr/) 2 and 12 ([#142])
+- Applied code formatting rules from [PSR-12](https://www.php-fig.org/psr/psr-12/) ([#142])
+- Resolved static analysis warnings reported by [PHPStan](https://phpstan.org/) &
+  [PHPMD](https://phpmd.org/) ([#143])
 
 ## Version 1.0.0 - 2021-09-05
 
@@ -223,3 +225,4 @@ and may require changes in applications that invoke these methods:_
 [#139]: https://github.com/hamstar/Wikimate/pull/139
 [#140]: https://github.com/hamstar/Wikimate/pull/140
 [#142]: https://github.com/hamstar/Wikimate/pull/142
+[#143]: https://github.com/hamstar/Wikimate/pull/143
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -61,6 +61,27 @@ class Wikimate
     protected $api;
 
     /**
+     * Default headers for Requests_Session
+     *
+     * @var array
+     */
+    protected $headers;
+
+    /**
+     * Default data for Requests_Session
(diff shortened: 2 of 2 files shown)


waldyrious (approved):
Nice! These changes all seem reasonable to me. The explicitness of the local variables in the call to `setSection()` was nice, but I guess the proper way to address this is to use [named parameters](https://devmount.medium.com/php-8-features-i-wish-also-existed-in-javascript-a4037230a1f0#def1).

Xymph:
> Nice! These changes all seem reasonable to me. The explicitness of the local variables in the call to `setSection()` was nice, but I guess the proper way to address this is to use [named parameters](https://devmount.medium.com/php-8-features-i-wish-also-existed-in-javascript-a4037230a1f0#def1).

True, but that was only introduced in PHP 8, while we don't want to ditch support for 5.3-7.x just for that reason.

Issue #22, opened 2014-06-12
Rewrite code to use rmccue/Requests


hamstar:
As discussed in #21, we need to remove the unmaintained Curl library written by shuber and instead use the [Requests](https://github.com/rmccue/Requests) library by rmccue.

We should use composer/packagist to add the dependency and then rewrite the code to use it.

We should really use [a facade](http://en.wikipedia.org/wiki/Facade_pattern) as well by making another class that deals with the Requests library (`RequestFacade`).  Then we only have to change the facade if we want to change the curl/request library again in the future.

PR #147, opened 2022-03-09
Bump Requests to Version2

nineff:


waldyrious:
Just for the record, I'm linking #22 and #25 where the adoption of Requests was discussed and implemented, respectively.

.gitignore
```diff
@@ -1,3 +1,3 @@
 wikimate_cookie.txt
-nbproject/private
+nbproject
```
Xymph, line 2: Good catch.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and [Keep a Changelog](http://keepachangelog.com/).
   [PHPMD](https://phpmd.org/) ([#143])
 - Clarified error response for edits denied by a CAPTCHA ([#145])
 - Activated default markdownlint rules ([#146])
+- Updated dependency on rmccue/requests to Version 2.x (fixes PHP 8.1 depcrecation warnings)
```
Xymph, line 17: Could you clarify those 8.1 deprecation warnings? According to the [Request doc site](https://requests.ryanmccue.info/), PHP 5.6 should still be supported in v2.0 (and [5.2 in 1.8](https://github.com/WordPress/Requests/tree/release/1.8.1)), so why would we need to update Wikimate?

Apart from that, it is of course good to stay with the times, but I didn't want to update to v2.0.0 because of potential migration pains at their end. v2.0.1 seems okay for Wikimate though, in my brief tests.

But, "depcrecation" is misspelled. And please add the PR number after the entry, with the corresponding link at the end, formatted like all other entries. Would be even nicer if you can rebase that together with commit 6437306.

Xymph (changes requested):

CHANGELOG.md
nineff: > Could you clarify those 8.1 deprecation warnings? According to the [Request doc site](https://requests.ryanmccue.info/), PHP 5.6 should still be supported in v2.0 (and [5.2 in 1.8](https://github.com/WordPress/Requests/tree/release/1.8.1)), so why would we need to update Wikimate?

Sure. The issue is not directly with Wikimate but with the requests V1.8 library. If you are on PHP 8.1, it causes several deprecation warnings due to changing return types.
e.g.: 
```
Deprecated: Return type of Requests_Cookie_Jar::offsetExists($key) should either be compatible with ArrayAccess::offsetExists(mixed $offset): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in [REDACTED]\vendor\rmccue\requests\library\Requests\Cookie\Jar.php on line 63

Call Stack:
    0.0002     361008   1. {main}() [REDACTED]\userpage.php:0
    0.0030     367600   2. Wikimate->__construct($api = '[REDACTED]', $headers = ???, $data = ???, $options = ???) [REDACTED]\userpage.php:12
    0.0030     368296   3. Wikimate->initRequests() [REDACTED]\vendor\hamstar\wikimate\Wikimate.php:149
    0.0034     368584   4. Requests_Session->__construct($url = '[REDACTED]', $headers = [], $data = [], $options = []) [REDACTED]\vendor\hamstar\wikimate\Wikimate.php:163
    0.0034     368584   5. Composer\Autoload\ClassLoader->loadClass($class = 'Requests_Cookie_Jar') [REDACTED]\vendor\rmccue\requests\library\Requests\Session.php:75
    0.0035     368744   6. Composer\Autoload\includeFile($file = '[REDACTED]\\vendor\\composer/../rmccue/requests/library\\Requests\\Cookie\\Jar.php') [REDACTED]\vendor\composer\ClassLoader.php:428
    0.0037     369488   7. include('[REDACTED]\vendor\rmccue\requests\library\Requests\Cookie\Jar.php') [REDACTED]\vendor\composer\ClassLoader.php:571
```
Updating Requests to Version 2.0.1 resolves those deprecations, but causes new warnings since Version 2 is namespaced with a backup for non-namespaced calls. I adjusted Wikimate to use the namespaced version in accordance with the [Requests Documentation](https://requests.ryanmccue.info/docs/upgrading.html). Then all Warnings are gone on PHP 8.1.

> But, "depcrecation" is misspelled. And please add the PR number after the entry, with the corresponding link at the end, formatted like all other entries. Would be even nicer if you can rebase that together with commit https://github.com/hamstar/Wikimate/commit/6437306b2d66c696d62969eadb8286a0f48a0a13.

Sorry, I'll fix the typo and the change on the `.gitignore`. But I'm not sure what you mean with the rebase, I'm still trying to learn git.

Xymph: > Sorry, I'll fix [...] the change on the .gitignore. 

Okay, according to [this advice](https://stackoverflow.com/a/1267580) only private NetBeans data needs to be excluded indeed.

But: at the end of Changelog, a link for 147 is also needed.

> But I'm not sure what you mean with the rebase, I'm still trying to learn git.

That's okay then.
@waldyrious Perhaps you can advise how to condense the commit entries before or after merging?

nineff: > But: at the end of Changelog, a link for 147 is also needed.
I did put that in as well

nineff:
Updated accordingly.

CHANGELOG.md
Xymph: > >  But: at the end of Changelog, a link for 147 is also needed.
> 
> I did put that in as well

Strange, at the end of the [raw file in the commit](https://raw.githubusercontent.com/hamstar/Wikimate/909815e2da7fa9daa6e04b6e92759056ff3bf1d8/CHANGELOG.md), there is no entry for 147.

nineff: Maybe you are viewing a cached version? Here's how it looks for me if I click your link:
grafik

Xymph (approved):

CHANGELOG.md
waldyrious: > @waldyrious Perhaps you can advise how to condense the commit entries before or after merging?

Assuming @nineff has granted Wikimate maintainers the permissions to edit this PR branch, we should be able to do the commit squashing ourselves. I'll try that right away :)

(For the record, the easiest option would be to simply merge the PR with a squash commit, but that would fail to conform to the semi-linear git graph [recommendation](https://github.com/hamstar/Wikimate/blob/master/GOVERNANCE.md#guidelines-for-pull-requests) from our governance guidelines.)

waldyrious:
Cool, I managed to update the branch after editing it locally with an interactive rebase.

@nineff, if you're starting with git, I recommend you give https://learngitbranching.js.org a try — it's an excellent resource to build intuition and confidence about git commands and what they're actually doing in the repository.

nineff:
> Cool, I managed to update the branch after editing it locally with an interactive rebase.
> 
> @nineff, thanks for the contribution! And by the way, if you're starting with git, I recommend you give https://learngitbranching.js.org a try — it's an excellent resource to build intuition and confidence about git commands and what they're actually doing in the repository.

Thanks, I'll have a look at it.

PR #148, opened 2022-05-22
Fix array key in case of login errors

nineff:
https://github.com/hamstar/Wikimate/blob/ecec68329821cdc5d72666f36eb5129347cc2a6f/Wikimate.php#L393 creates $error['auth'] instead of $error['login'] key. Adjusted example accordingly


Xymph:
Yup, done. Thanks.

PR #150, opened 2023-03-29
Improved more PHPDoc comments

Xymph:
This resolves inline @see tags not getting linked in PHPDoc blocks, and various styling issues where the [summary line](https://docs.phpdoc.org/guide/guides/docblocks.html#summary) wasn't properly separated from the description section. This makes every (I hope) docblock look good in phpDocumentor.
Also includes grammar improvements.

Hmm, should the two "improve" PRs be in the Changed section of the changelog, instead?

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -858,7 +858,7 @@ class WikiPage
 
     /**
      * Constructs a WikiPage object from the title given
-     * and associate with the passed Wikimate object.
+     * and associates it with the passed Wikimate object.
      *
      * @param  string    $title     Name of the wiki article
      * @param  Wikimate  $wikimate  Wikimate object
@@ -901,9 +901,10 @@ class WikiPage
     }
 
     /**
-     * Returns an array sections with the section name as the key
-     * and the text as the element, e.g.
+     * Returns an array of sections with the section name as the key
+     * and the text as the element.
      *
+     * For example:
      * array(
      *   'intro' => 'this text is the introduction',
      *   'History' => 'this is text under the history section'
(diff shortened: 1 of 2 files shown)


Wikimate.php
```diff
@@ -2075,7 +2088,7 @@ public function getUser($revision = null)
 
     /**
      * Returns the ID of the user who uploaded this file,
-     * or of its specified revision.
```
waldyrious, line 2078: How about "...or a given revision of it, if specified"? (Same comment for line 2055).

waldyrious:
> Hmm, should the two "improve" PRs be in the Changed section of the changelog, instead?

Nah, they do fix (documentation) issues — they're not particularly serious, as they don't affect the behavior of the software itself, but they were corrections nevertheless. I'd say we keep them in the "Fixed" section.

Wikimate.php
Xymph: Uhm, that phrase occurs 19 times in total. I agree your proposal is more precise, but then it would apply to a lot more than 2 lines. Is that what you actually intended?

waldyrious: I'd prefer that, yes. I didn't realize it was so common a phrase, but since you're already making changes across the board in the docs, I don't feel like it's expanding the scope of this PR too much.

Xymph: Done. Didn't rebase, the commit can standalone.

waldyrious: Agreed.

waldyrious (approved):
LGTM!

PR #151, opened 2023-05-10
Added internal mechanism to handle version-dependent parameters of API modules

Xymph:
Many API modules accept new parameters (or deprecate old ones) as of specific MediaWiki versions. The [paraminfo action](https://www.mediawiki.org/wiki/API:Parameter_information) allows to obtain the list of supported parameters of one or more modules in the MediaWiki instance that Wikimate is used with.

This PR adds an internal mechanism to obtain/store/use supported parameters with the API calls used by Wikimate. It is initially used in `Wikimate::logout()` so that this no longer encounters a warning before MW v1.34.

### DIFF
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -473,17 +473,18 @@ class Wikimate
      */
     public function logout()
     {
-        // Obtain logout token first
-        if (($logouttoken = $this->token()) === null) {
-            return false;
-        }
-
-        // Token is needed in MediaWiki v1.34+, older versions produce an
-        // 'Unrecognized parameter' warning which can be ignored
         $details = array(
             'action' => 'logout',
-            'token' => $logouttoken,
         );
+
+        // Check if token parameter is required (it is since MediaWiki v1.34)
+        if ($this->supportsModuleParam('logout', 'token')) {
+            // Obtain logout token
+            if (($logouttoken = $this->token()) === null) {
+                return false;
+            }
+            $details['token'] = $logouttoken;
+        }
 
         // Send the logout request
         $logoutResult = $this->request($details, array(), true);
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -153,6 +153,14 @@ class Wikimate
     private $csrfToken = null;
 
     /**
+     * Stored parameter lists for API modules
+     *
+     * @var array
+     * @link https://www.mediawiki.org/wiki/Special:MyLanguage/API:Parameter_information
+     */
+    private $moduleParams = array();
+
+    /**
      * Creates a new Wikimate object.
      *
      * @param   string    $api      Base URL for the API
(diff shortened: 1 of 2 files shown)


Xymph:
FYI, I found that `Wikimate::supportsModuleParam()` needs to be public. A protected method does not work for classes WikiPage/WikiFile, because those are not child classes, but get passed the Wikimate object to their constructors for storage in `$this->wikimate`.

Xymph:
> FYI, I found that `Wikimate::supportsModuleParam()` needs to be public. A protected method does not work for classes WikiPage/WikiFile, because those are not child classes, but get passed the Wikimate object to their constructors for storage in `$this->wikimate`.

Ok, sticking with this approach. For an internal mechanism, having the method public (and overloadable) is not really desirable. But if it remains private, then the decision logic whether to add the `deletetalk` parameter to `delete()` requests would need to be moved away from the WikiPage/WikiFile classes into `Wikimate::delete()`. And other action modules with version-dependent parameters could also have logic split between the classes. That is even less desirable (even though I don't foresee coverage of such modules in Wikimate for a long time).

Wikimate.php
```diff
@@ -346,6 +354,57 @@ protected function token($type = self::TOKEN_DEFAULT)
         }
     }
 
+    /**
+     * Obtains parameter lists for API modules.
+     * For specific API modules where supported parameters vary with
+     * MediaWiki versions, fetch and store these parameter lists for
+     * use by the API calls of those modules.  Current modules list:
+     * - logout
```
waldyrious, line 362:
```suggestion
     * Obtains parameter lists for API modules.
     * For specific API modules where supported parameters vary with
     * MediaWiki versions, this method is used to fetch and store
     * these parameter lists for use by the API calls of those modules.
     * Current list of methods that make use of this:
     * - logout
```

Wikimate.php
```diff
@@ -410,18 +472,19 @@ public function login($username, $password, $domain = null)
...
-        }
-
-        // Token is needed in MediaWiki v1.34+, older versions produce an
-        // 'Unrecognized parameter' warning which can be ignored
         $details = array(
             'action' => 'logout',
-            'token' => $logouttoken,
         );
 
+        // Check if token parameter is required (since MediaWiki v1.34)
```
waldyrious, line 479:
```suggestion
        // Check if token parameter is required (it is since MediaWiki v1.34)
```

waldyrious (commented):
Apologies for the (very) late review! I think this looks pretty good. The only suggestions I left were for the documentation, but I may have misinterpreted some parts, so feel free to push back if any of the changes don't seem reasonable to you.

Wikimate.php
Xymph: I worded the last sentence slightly different: "Current list of modules that employ this mechanism:"

Xymph:
Thanks, changes made and rebased.

Wikimate.php
waldyrious: My question is whether we want to refer to (Wikimate) methods or (MediaWiki API) modules. It seemed that the former makes more sense, but maybe I'm misreading your intention?

waldyrious: Got it. Do you think it might be worth linking "logout" to https://www.mediawiki.org/wiki/API:Logout?

And also, how about rephrasing slightly to "Current list of modules for which this mechanism is employed"? I'm asking because who _employs_ the mechanism is Wikimate, while the API _provides_ the mechanism for use in all modules (IIUC).

waldyrious (approved):
LGTM! Great work as always :)

Xymph:
Phew, finally. :smiley:   Thanks.

PR #152, opened 2023-07-20
Support deletetalk parameter in WikiPage::delete() and WikiFile::delete()

Xymph:
This adds support for deleting corresponding talk pages when deleting pages and files (which was introduced in MediaWiki v1.38), using the version-dependent mechanism added in #151 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and [Keep a Changelog](http://keepachangelog.com/).
 ### Added
 
 - Added internal mechanism to handle version-dependent parameters of API modules ([#151])
+- Support version-dependent `deletetalk` parameter in `WikiPage::delete()` and `WikiFile::delete()` ([#152])
 
 ### Changed
 
@@ -23,6 +24,8 @@ and [Keep a Changelog](http://keepachangelog.com/).
 ### Fixed
 
 - Fixed example not using correct array key for invalid login errors ([#148])
+- Prevented warning in `Wikimate::logout()` about `token` parameter that was unsupported
+  before MediaWiki v1.34 ([#151])
 - Improved various code comments ([#149])
 - Improved more PHPDoc comments ([#150])
 
@@ -246,3 +249,4 @@ and may require changes in applications that invoke these methods:_
 [#149]: https://github.com/hamstar/Wikimate/pull/149
 [#150]: https://github.com/hamstar/Wikimate/pull/150
 [#151]: https://github.com/hamstar/Wikimate/pull/151
+[#152]: https://github.com/hamstar/Wikimate/pull/152
--- a/Wikimate.php
+++ b/Wikimate.php
@@ -361,6 +361,7 @@ class Wikimate
      * these parameter lists for use by the API calls of those modules.
      * Current list of modules for which this mechanism is employed:
      * - {@link https://www.mediawiki.org/wiki/Special:MyLanguage/API:Logout logout}
+     * - {@link https://www.mediawiki.org/wiki/Special:MyLanguage/API:Delete delete}
      *
      * @return void
      * @link https://www.mediawiki.org/wiki/Special:MyLanguage/API:Parameter_information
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and [Keep a Changelog](http://keepachangelog.com/).
 ### Added
 
 - Added internal mechanism to handle version-dependent parameters of API modules ([#151])
+- Support version-dependent `deletetalk` parameter in `WikiPage::delete()` and `WikiFile::delete()` ([#152])
```
waldyrious, line 13: Shouldn't there also be a similar changelog entry for the logout method? :thinking:

Wikimate.php
```diff
@@ -1419,10 +1420,11 @@ public function newSection($name, $text)
...
      * Deletes the page.
      *
-     * @param   string   $reason  Reason for the deletion
-     * @return  boolean           True if page was deleted successfully
+     * @param   string   $reason   Reason for the deletion
+     * @param   boolean  $talktoo  True to delete talk page too (MW 1.38+)
+     * @return  boolean            True if page was deleted successfully
      */
-    public function delete($reason = null)
+    public function delete($reason = null, $talktoo = false)
```
waldyrious, line 1427: I find it odd that [the default in MediaWiki is false](https://github.com/wikimedia/mediawiki/blob/1.38.0/includes/api/ApiDelete.php#L175). Would you be open to deviate from that behavior and default to deleting the talk page too? I can't think of many cases where one would want to delete the main page but keep its talk page around, but maybe I just lack imagination! :sweat_smile:

Wikimate.php
```diff
@@ -1432,6 +1434,10 @@ public function delete($reason = null)
         if (!is_null($reason)) {
             $data['reason'] = $reason;
         }
+        // Check if deletetalk parameter is supported (it is since MediaWiki v1.38)
```
waldyrious, line 1437:
```suggestion
        // Check if the `deletetalk` parameter is supported (it is since MediaWiki v1.38)
```

(Does markdown code formatting work in the rendered docs?)

Wikimate.php
```diff
@@ -2334,6 +2341,10 @@ public function delete($reason = null, $archivename = null)
         if (!is_null($archivename)) {
             $data['oldimage'] = $archivename;
         }
+        // Check if deletetalk parameter is supported (it is since MediaWiki v1.38)
```
waldyrious, line 2344:
```suggestion
        // Check if the `deletetalk` parameter is supported (it is since MediaWiki v1.38)
```

Wikimate.php
Xymph: Uhm, inline comments are not included in the rendered docs, so the backticks are not useful. And to keep the line length close to 80 characters (for undeletetalk in the next PR too), I think it's fine to leave out "the" here. It's just a comment after all. :)

CHANGELOG.md
Xymph: Oh, right. Added.

Wikimate.php
Xymph: Historically talk pages were not deleted while deleting a main/file space page, this required a separate talk delete, so the default makes sense. And a common situation (at least on the Doom Wiki) is that the talk page has a discussion/vote about the article deletion, which is then preserved to document the voting result. So I definitely don't think defaulting to `true` is a good angle for Wikimate.

Wikimate.php
waldyrious: Uh, silly me, I clearly wasn't paying attention. Ok, fair enough, let's disregard this suggestion.

Wikimate.php
waldyrious: Ah, that does make sense, yeah. Ok then, let's follow suit then.

CHANGELOG.md
```diff
@@ -23,6 +24,8 @@ and [Keep a Changelog](http://keepachangelog.com/).
 ### Fixed
 
 - Fixed example not using correct array key for invalid login errors ([#148])
+- Prevented warning in `Wikimate::logout()` about `token` parameter that was unsupported
+  before MediaWiki v1.34 ([#151])
```
waldyrious, line 28: I'd have preferred this to be in a separate commit (ideally it would have been part of the previous PR, but that ship has sailed). However, it's such a minor thing that I don't think it's worth the hassle to address now.

waldyrious (approved):
LGTM!

PR #154, opened 2023-07-21
Fix markdownlint execution in the CI pipeline

waldyrious:
Follow up from the [conversation from #153](https://github.com/hamstar/Wikimate/pull/153#issuecomment-1644389364).

The advanced glob expression we were using needs to be quoted, as explained in <https://github.com/igorshubovych/markdownlint-cli#globbing>:

> `markdownlint-cli` supports advanced globbing patterns like `**/*.md`. With shells like Bash, it may be necessary to quote globs so they are not interpreted by the shell.

### DIFF
--- a/.github/workflows/pull-request-checks.yml
+++ b/.github/workflows/pull-request-checks.yml
@@ -1,17 +1,22 @@
 name: "Pull Request Checks"
 on: pull_request
 jobs:
-  # Enforce the update of the changelog file
+  # Ensure that the changelog file has been updated in this PR
   check-changelog-change:
     runs-on: ubuntu-latest
     steps:
     - uses: dangoslen/changelog-enforcer@v3
-  # Ensure markdown files are formatted consistently
+  # Ensure that markdown files are formatted consistently (according to .markdownlint.yml)
   lint-markdown-files:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: xt0rted/markdownlint-problem-matcher@v2
-      - run: npm install -g markdownlint-cli
-      - run: npx --silent glob '**/*.md'
-      - run: markdownlint '**/*.md'
+      - name: Fetch the Wikimate code into the Actions runner
+        uses: actions/checkout@v3
+      - name: Install the markdownlint-problem-matcher action
+        uses: xt0rted/markdownlint-problem-matcher@v2
+      - name: Install markdownlint-cli
+        run: npm install -g markdownlint-cli
+      - name: List files that will be processed
+        run: npx --silent glob '**/*.md'
+      - name: Run markdownlint
+        run: markdownlint '**/*.md'
--- a/.github/workflows/pull-request-checks.yml
+++ b/.github/workflows/pull-request-checks.yml
@@ -13,4 +13,4 @@ jobs:
       - uses: actions/checkout@v3
       - uses: xt0rted/markdownlint-problem-matcher@v2
       - run: npm install -g markdownlint-cli
-      - run: markdownlint **/*.md
+      - run: markdownlint '**/*.md'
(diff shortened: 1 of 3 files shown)


waldyrious:
Grr, the checks are not running for some reason :(

Xymph:
> Grr, the checks are not running for some reason :(

One still isn't. Formatting the changelog PR reference with [] instead of an actual link at the end of the file seems to cause the error, and also deviates from what we've done until now. Try with the old mechanism?

waldyrious:
That's deliberate, actually. It turns out that markdownlint is not able to detect the missing link target in [reference links](https://spec.commonmark.org/0.30/#reference-link) when we use the [shortcut syntax](https://spec.commonmark.org/0.30/#shortcut-reference-link) `[foo]`, but it can when we use the [collapsed syntax](https://spec.commonmark.org/0.30/#collapsed-reference-link) `[foo][]`. I have opened https://github.com/DavidAnson/markdownlint/issues/915 to ask for that feature, but until that's implemented and released, we need to resort to using the collapsed syntax to make the links be processed.

I'm no longer blocked but I ran out of time. :sweat_smile: I'll get back to experimenting with this PR ASAP.

Xymph:
Well, in that case, I think I'll go ahead to prepare v1.1 with the missing link, and this PR can be wrapped up later. It doesn't need to hold up the release.

But no longer having to expand the long link list in the future would be nice indeed. I'm just wondering whether the collapsed link will find the PRs in the original repo, even when viewing the changelog in our forks?

waldyrious:
> Well, in that case, I think I'll go ahead to prepare v1.1 with the missing link, and this PR can be wrapped up later. It doesn't need to hold up the release.

Why keep the missing link in the release? I would fix the link which we have already detected as missing, since that is orthogonal to the automated detection of future missing links.

> But no longer having to expand the long link list in the future would be nice indeed. I'm just wondering whether the collapsed link will find the PRs in the original repo, even when viewing the changelog in our forks?

Oh, sorry, I'm afraid I gave you the wrong impression. The list at the end of the file still needs to be there regardless of whether we use the `[foo]` or the `[foo][]` syntax for the links; the only difference would be to have markdownlint warn us when one of the entries there is missing, so that the same issue won't go unnoticed again.

Xymph:
My bad wording, I meant: "prepare v1.1 _while adding_ the missing link"

Oh, okay, I had the impression the new syntax would magically link the PR somehow. Carry on then. :smile:

waldyrious:
Ok, I think I've got it running. Now we just need to wait for a new release of [markdownlint-cli](https://github.com/igorshubovych/markdownlint-cli) or [markdownlink-cli2](https://github.com/DavidAnson/markdownlint-cli2) that includes support for [markdownlint 0.31.0](https://github.com/DavidAnson/markdownlint/releases/tag/v0.31.0) (the version that introduced support for the shortcut syntax checking I asked for as [mentioned above](https://github.com/hamstar/Wikimate/pull/154#issuecomment-1646361358) — thanks @DavidAnson!).

Afterwards I'll clean up the commits in this PR so it'll be ready to merge.

DavidAnson:
Soon!

waldyrious:
Alright! Now that markdownlint-cli has gotten a [new release](https://github.com/igorshubovych/markdownlint-cli/releases/tag/v0.37.0) that includes markdownlint 0.31.0 (and therefore supports the new `shortcut_syntax` parameter for the MD052 rule), the `npm install` command in the GitHub Actions workflow will automatically get the new version; so we are now able to add that parameter to our `.markdownlint.yml` file (which I did in df4116468) and get warnings for undefined reference link targets.

I have confirmed that an undefined URL reference does produce a markdownlint violation in the CI with the new config option. After that test I reverted the violation, naturally. I also cleaned up the commits and renamed the PR to be more accurate.

@Xymph this should now be ready to merge — please take a look at your convenience!

Xymph:
Lookin' good. :+1:

PR #155, opened 2023-07-23
Describe usage of Wikimate::undelete()

Xymph:
PR #153 was even hastier than thought before: I also overlooked that I had prepared an undelete section with example calls for USAGE.md. So this PR adds it and the previous missing changelog link.


waldyrious (approved):
LGTM! Sorry for the late review.

PR #162, opened 2023-09-11
Move reference link URLs in CHANGELOG.md to their sections

waldyrious:
Just an experimental change that may make this file easier to edit (and to consume, in source mode).


Xymph (approved):
Works for me.

PR #163, opened 2024-02-22
Update API URLs to typical path, update other links, and use https

Xymph:
I noticed a few http links now redirect to https, and an extended URL scan pointed to some other links also worth updating.

### DIFF
--- a/.markdownlint.yml
+++ b/.markdownlint.yml
@@ -21,7 +21,7 @@ MD013:
 # MD024/no-duplicate-heading
 MD024:
   # Only check sibling headings
-  allow_different_nesting: true
+  siblings_only: true
 
 # MD026/no-trailing-punctuation
 MD026:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,15 @@
 
 This page lists the changes that were done in each version of Wikimate.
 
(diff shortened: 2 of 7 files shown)


Xymph:
Hmm, looks like markdownlint needs to be taught how to accept our changelog.

Xymph:
Hey @waldyrious, could you please resolve this GH action problem? Thanks.

waldyrious:
Hmm, it seems they [changed](https://github.com/DavidAnson/markdownlint/commit/f2725178b1e83b07d7f7cc2db64d42f6aa14b8a1) the configuration of MD024 to only accept `siblings_only` rather than `allow_different_nesting`, which is what we used.

We need to change this code:

https://github.com/hamstar/Wikimate/blob/e6d3fdac49ab87f18e96c800658ce32d5eb5b388/.markdownlint.yml#L21-L24

```diff
 # MD024/no-duplicate-heading
 MD024:
   # Only check sibling headings
-  allow_different_nesting: true
+  siblings_only: true
```

Do you want to do that in this PR, or should we make it a separate one?

Xymph:
Welcome back. The clean way would be a separate PR, tbh.

Xymph:
@waldyrious The silence has been deafening long enough. :stuck_out_tongue_winking_eye: 

I fixed the markdown rule in this PR and merged it per the Governance rule 2 exception.

